PR #23, opened 2014-08-31
connect() method can use NEOVIM_LISTEN_ADDRESS 

kryskool:
Hi @tarruda 

Thanks for this library, i just improve the possibility to use the NEOVIM_LISTEN_ADDRESS in connect() method if address is empty, with this improvement we can directly execute python script in neovim like (neovim set NEOVIM_LISTEN_ADDRESS if not exists)

```
:! python myscript.py
```

Regards,

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,14 @@ Open the python REPL with another terminal connect to Neovim:
 [1, 2, 3]
 ```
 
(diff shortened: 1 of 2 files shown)


justinmk:
Not really sure this is a feature we want, and the PR hasn't been updated. So, closing.

PR #72, opened 2014-12-21
[WIP/RFC] allow polling on a provided fd

bfredl:
This allows a plugin to poll for readability and/or writability on a specified fd. 
[here](https://gist.github.com/bfredl/5f90c9ebca958401b4ee) is an example integrating with a zeromq socket.

However, I'm not actually sure we want to spawn a greenlet for every poll event, because this event doesn't imply that there actually is useful data that needs to be handled. ( For instance, `zmq` uses the same fd also for internal communication with its worker threads and so at times there could be lots of spurious poll events) Perhaps its better to spawn a greenlet only when there is message, for instance:

```
     def _on_poll(self, fd, read, write):
        while self.sock.getsockopt(zmq.EVENTS) & zmq.POLLIN:
            self.vim.session.schedule(lambda: self.on_msg( self.sock.recv()))
```

One might even want back to yield back to an existing greenlet that is blocking on receiving (or sending) on the fd. 
Or it could be that spawning new greenlets is so cheap, that its not worth bothering avoiding it, I don't know.

### DIFF
--- a/neovim/api/nvim.py
+++ b/neovim/api/nvim.py
@@ -322,6 +322,20 @@ class Nvim(object):
         """Return new src_id for use with Buffer.add_highlight."""
         return self.current.buffer.add_highlight("", 0, src_id=0)
 
+    def _error_wrapper(self, fn, call_point, *args, **kwargs):
+        if fn is None:
+            return None
+        def handler():
+            try:
+                fn(*args, **kwargs)
+            except Exception as err:
+                msg = ("error caught while executing async callback:\n"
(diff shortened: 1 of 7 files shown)


bfredl:
Updated, now it has simpler `vim.session.poll_fd(fd, on_readable=callback)` api, and a greenlet is not automatically spawned.

bfredl:
Updated. Goes back to spawn a greenlet per default, to behave like a an ordinary (async) neovim command handler or `async_call` callback (i e api calls can be done in the poll callback, principle of least surprise).

PR #74, opened 2015-01-02
[RDY] UI prototypes

tarruda:
Prototyping a graphical and text UI for helping me fix issues with the abstract ui code. This branch depends on neovim/neovim#1657.

### DIFF
--- a/neovim/ui/screen.py
+++ b/neovim/ui/screen.py
@@ -1,0 +1,134 @@
+"""Common code for graphical and text UIs."""
+from ..compat import IS_PYTHON3
+
+
+__all__ = ('Screen',)
+
+
+if not IS_PYTHON3:
+    range = xrange  # NOQA
+
+
+class Cell(object):
+    def __init__(self):
+        self.text = ' '
+        self.attrs = None
+
+    def __repr__(self):
+        return self.text
+
+    def get(self):
+        return self.text, self.attrs
+
+    def set(self, text, attrs):
+        self.text = text
+        self.attrs = attrs
+
+    def copy(self, other):
+        other.text = self.text
+        other.attrs = self.attrs
+
+
+class Screen(object):
+
+    """Store nvim screen state."""
+
+    def __init__(self, columns, rows):
+        """Initialize the Screen instance."""
+        self.columns = columns
+        self.rows = rows
+        self.row = 0
+        self.col = 0
+        self.top = 0
+        self.bot = rows - 1
+        self.left = 0
+        self.right = columns - 1
+        self._cells = [[Cell() for c in range(columns)] for r in range(rows)]
+
+    def clear(self):
+        """Clear the screen."""
+        self._clear_region(self.top, self.bot, self.left, self.right)
+
+    def eol_clear(self):
+        """Clear from the cursor position to the end of the scroll region."""
+        self._clear_region(self.row, self.row, self.col, self.right)
+
+    def cursor_goto(self, row, col):
+        """Change the virtual cursor position."""
+        self.row = row
+        self.col = col
+
+    def set_scroll_region(self, top, bot, left, right):
+        """Change scroll region."""
+        self.top = top
+        self.bot = bot
+        self.left = left
+        self.right = right
+
+    def scroll(self, count):
+        """Shift scroll region."""
+        top, bot = self.top, self.bot
+        left, right = self.left, self.right
+        if count > 0:
+            start = top
+            stop = bot - count + 1
+            step = 1
+        else:
+            start = bot
+            stop = top - count - 1
+            step = -1
+        # shift the cells
+        for row in range(start, stop, step):
+            target_row = self._cells[row]
+            source_row = self._cells[row + count]
+            for col in range(left, right + 1):
+                tgt = target_row[col]
+                source_row[col].copy(tgt)
+        # clear invalid cells
+        for row in range(stop, stop + count, step):
+            self._clear_region(row, row, left, right)
+
+    def put(self, text, attrs):
+        """Put character on virtual cursor position."""
+        cell = self._cells[self.row][self.col]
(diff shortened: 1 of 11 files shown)


bfredl:
Nice, input (including altGr) works very well. A slight discrepancy from teminal nvim though: clicking any word will visually select the entire word, unlike the mouse in the terminal which will just move the cursor in normal mode. (It looks like every press becomes a doubleclick)

tarruda:
>  A slight discrepancy from teminal nvim though: clicking any word will visually select the entire word, unlike the mouse in the terminal which will just move the cursor in normal mode. (It looks like every press becomes a doubleclick)

Must be some timing issue, I just tried double clicking with a relatively high interval and it selects the word, like in terminal vim. FWIW the behavior is tested [here](https://github.com/neovim/neovim/blob/master/test/functional/ui/mouse_spec.lua#L94-L103)

tarruda:
@bfredl I'm done with the gtk UI, do you mind giving it a try and check for pending issues? The mouse click issue should be fixed now.

Now I'm gonna prototype a libtickit UI using cffi, and then translate it to C in neovim core.

bfredl:
The issue of missing formatting I had before is now gone, but I still have the mouse-click issue. (This also doesn't support external input methods yet, but I have a [branch locally](https://github.com/bfredl/python-client/commit/2425fe9f32e3f7d3702370cdfc5803fe293a90b8) that mostly implements this. I will push PR when this is done)

tarruda:
Done with this PR, I will now perform some final adjustments in neovim/neovim#1657 and then merge everything.

neovim/ui/screen.py
```diff
@@ -0,0 +1,138 @@
...
+    def clear(self):
+        """Clear the screen."""
+        self._clear_region(0, self.rows - 1, 0, self.columns - 1)
+
+    def eol_clear(self):
+        """Clear from the cursor position to the end of the scroll region."""
+        try:
+            self._clear_region(self.row, self.row, self.col, self.right)
+        except IndexError:
+            print self.row, self.col, self.right
```
fwalch: Braces are missing here (for Python 3).

tarruda: This is just a debug print, nvim should not send invalid indexes. Removed now

Tranquility:
It still happens. I tried to locate the problem for hours. I even tried to break your vimrc (which worked fine). Please try mine and see whether you can replicate the problem. https://gist.github.com/Tranquility/6074b37b33196a2ee26c

tarruda:
@Tranquility I've fixed the tickit UI to render default foreground/background properly, but you need to update this and neovim/neovim#1657

Tranquility:
The cmd-line bg color is still wrong and also if I open a file directly the part where no text is has a wrong background. 
neovim-issue

If I call `:redraw!` after opening a file everything works.  But if I open pynvim without a file, open a file afterwards and call then `:redraw!` I get the previously described ui glitch.
neovim-ui

tarruda:
> The cmd-line bg color is still wrong and also if I open a file directly the part where no text is has a wrong background. 

If you are using [this colorscheme](https://github.com/w0ng/vim-hybrid/blob/master/colors/hybrid.vim), it won't work without modification because tests like `if t_Co` or `if has('gui_running')` don't make sense in this new UI architecture(it is possible to have multiple UIs attached, and they can be TUIs or GUIs)

> But if I open pynvim without a file, open a file afterwards and call then :redraw! I get the previously described ui glitch.

I'm getting a lot of problems with your vimrc and color scheme, will investigate tomorrow

fwalch:
> the part where no text is has a wrong background

I get this with `haya14busa/incsearch.vim`. To reproduce, setting up the plugin, adding `map /  <Plug>(incsearch-forward)` to the `.nvimrc` and typing `/` should be enough.

62724644-5f1e-400a-b66f-6de942bb718c

tarruda:
@Tranquility do you get redraw glitches if ultisnips is disabled?

Tranquility:
@tarruda only disabling ultisnips was not enough but after disabling ultisnips AND youcompleteme the problem seems to be gone.

Tranquility:
@tarruda everything works now!

tarruda:
> @tarruda everything works now!

Even the background issue?

Tranquility:
Ok, I found something else. Maybe it is to early for that but the shell command output is very slow and I can't terminate it with ctrl-c.

I found that out by using fugitives :Ggrep with a pattern that has a lot of matches.

Tranquility:
> Even the background issue?

Yes, that is fixed. Only the command line bg color is wrong before the first redraw.

Tranquility:
Ok, looks like it's already tracked by https://github.com/neovim/neovim/issues/1234? But it still takes a really long time to display the output.

tarruda:
@Tranquility did the last commit improve command output speed for you?

Tranquility:
@tarruda no it is still very slow. I just execute :Glog in the neovim folder and it takes more than 10 seconds.

tarruda:
> @tarruda no it is still very slow. I just execute :Glog in the neovim folder and it takes more than 10 seconds.

Unfortunately there's not much I can do right now, the bottleneck is in the internal scroll function which is implemented in pure python. Here are the initial entries of the profiler output:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2679   13.720    0.005   26.644    0.010 screen.py:68(scroll)
 19237899   12.514    0.000   12.514    0.000 screen.py:27(copy)
       37    4.524    0.122    4.524    0.122 {method 'poll' of 'select.epoll' objects}
      233    2.231    0.010    2.231    0.010 {_cffi__xeb0d4df1xe5c9a0b5.tickit_term_flush}
    86749    1.388    0.000    1.646    0.000 screen.py:105(iter)
 ```

It should run much faster once its ported to C

PR #76, opened 2015-01-15
Try to prevent duplicated mouse clicks.

fmoralesc:
Note: releases are still duplicated.


bfredl:
Still  an issue, but an issue in `neovim/python-gui` now.

PR #77, opened 2015-01-18
Python3 fixes: functools.partial and encoding

equalsraf:
- Partial fix for neovim/neovim#1829
- I'm a bit at lost as to why the functools.partial is being used here, It
  clearly does not work for Python3, but it is not clear to me what is it
  supposed to be doing @tarruda?
- Encoding names in Python3 need to be strings, fixes Host._nvim_encoding
- Binary strings from msgpack are decoded as the bytes type, but rpc method
  names used by rpc_export are unicode strings, this causes a comparison
  between the request names and the exported functions to fail (use &encoding
  to decode)

### DIFF
--- a/neovim/plugin/host.py
+++ b/neovim/plugin/host.py
@@ -37,6 +37,8 @@ class Host(object):
             'shutdown': self.shutdown
         }
         self._nvim_encoding = nvim.options['encoding']
+        if IS_PYTHON3 and isinstance(self._nvim_encoding, bytes):
+            self._nvim_encoding = self._nvim_encoding.decode('ascii')
 
     def start(self, plugins):
         """Start listening for msgpack-rpc requests and notifications."""
@@ -51,6 +53,8 @@ class Host(object):
 
     def _on_request(self, name, args):
         """Handle a msgpack-rpc request."""
+        if IS_PYTHON3 and isinstance(name, bytes):
+            name = name.decode(self._nvim_encoding)
         handler = self._request_handlers.get(name, None)
         if not handler:
             msg = 'no request handler registered for "%s"' % name
@@ -64,6 +68,8 @@ class Host(object):
 
     def _on_notification(self, name, args):
         """Handle a msgpack-rpc notification."""
+        if IS_PYTHON3 and isinstance(name, bytes):
+            name = name.decode(self._nvim_encoding)
         handler = self._notification_handlers.get(name, None)
         if not handler:
             warn('no notification handler registered for "%s"', name)
--- a/neovim/plugin/decorators.py
+++ b/neovim/plugin/decorators.py
@@ -1,8 +1,10 @@
 """Decorators used by python host plugin system."""
+
 import inspect
-
-
 import logging
+
+from ..compat import IS_PYTHON3
+
 logger = logging.getLogger(__name__)
 debug, info, warn = (logger.debug, logger.info, logger.warn,)
 __all__ = ('plugin', 'rpc_export', 'command', 'autocmd', 'function',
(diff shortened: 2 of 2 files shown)


tarruda:
> I'm a bit at lost as to why the functools.partial is being used here, It clearly does not work for Python3, but it is not clear to me what is it supposed to be doing @tarruda?

For plugins that are not classes, the functions need to receive the Nvim instance as first argument. The functools.partial is there to bind a Nvim instance configured for a plugin handler(takes into consideration encoding decorators for example.

To be python3 compatible, it could be replaced by
```python
nvim = self._configure_nvim_for(fn)
fn2 = lambda *args: fn(nvim, *args)
```

tarruda:
LGTM, do you want to make any further changes before I merge?

equalsraf:
> For plugins that are not classes, the functions need to receive the Nvim instance as first argument. The functools.partial is there to bind a Nvim instance configured for a plugin handler(takes into consideration encoding decorators for example.

Makes sense, but in python3 it is binding the argument even when the plugin is a class. It seems the @plugin decorator is not working as intended - `inspect.getmembers` returns an empty list and _nvim_bind is never set to False (i.e. inspect considers the members as python functions instead of class methods this seems to be part of an open [issue](bugs.python.org/issue16851) for Python3).

I'm thinking about replacing the predicate with just a check for `_nvim_bind` - it seems a bit too pedantic to check for _nvim_bind in a method, inside a decorator that should only be used for classes.

```diff
diff --git a/neovim/plugin/decorators.py b/neovim/plugin/decorators.py
index 083146b..bb77cca 100644
--- a/neovim/plugin/decorators.py
+++ b/neovim/plugin/decorators.py
@@ -1,5 +1,6 @@
 """Decorators used by python host plugin system."""
 import inspect
+from ..compat import IS_PYTHON3
 
 
 import logging
@@ -20,8 +21,11 @@ def plugin(cls):
     # decorated functions have a bound Nvim instance as first argument.
     # For methods in a plugin-decorated class this is not required, because
     # the class initializer will already receive the nvim object.
-    for _, fn in inspect.getmembers(cls, inspect.ismethod):
-        if hasattr(fn, '_nvim_bind'):
+    predicate = lambda fn: hasattr(fn, '_nvim_bind')
+    for _, fn in inspect.getmembers(cls, predicate):
+        if IS_PYTHON3:
+            fn._nvim_bind = False
+        else:
             fn.im_func._nvim_bind = False
     return cls
```

The last piece of the puzzle is `im_func` needed in Python2 but not Python3?. But with that last diff it is working for both versions.

Shougo:
Can you merge the Pull Request for Python3?

tarruda:
Sure, I was just waiting for confirmation that its ready but I'm going to assume it is

neovim/plugin/host.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, nvim):
             'shutdown': self.shutdown
         }
         self._nvim_encoding = nvim.options['encoding']
+        if IS_PYTHON3:
+            self._nvim_encoding = self._nvim_encoding.decode('ascii')
```
tarruda: Isn't it possible for `self._nvim_encoding` to already be a string(in which case the "decode" will fail)

equalsraf: Sorry, I've been away. I'll fix this.

equalsraf:
Pushed a fix, scrutinizer failed, it seems it timed out.

Shougo:
:+1:

tarruda:
thanks

tarruda:
I've pushed a new release with these fixes

Issue #81, opened 2015-02-04
os.fchdir not available in Windows


equalsraf:
os.fchdir imported in [neovim/api/nvim.py](https://github.com/neovim/python-client/blob/4eadb632c69f3d030122fc29a3f9517b2d68cc79/neovim/api/nvim.py#L18) is not available in Windows. import neovim fails in Windows 

```python
  File "c:\Python27\lib\site-packages\neovim\api\nvim.py", line 18, in <module>
    os_fchdir = os.fchdir
AttributeError: 'module' object has no attribute 'fchdir'
```

I couldn't find any reference to the method in other parts of the code, is it really necessary?

tarruda:
> I couldn't find any reference to the method in other parts of the code, is it really necessary?

Not really, feel free to remove it. It was just added for compatibility with the old vim-python API(the compatibility layer has moved to its own [module](https://github.com/neovim/neovim/blob/master/runtime/autoload/provider/script_host.py) in the main repo)

justinmk:
closed by #82

PR #82, opened 2015-02-04
Remove use of os.fchdir

equalsraf:
- fchdir is not portable to Windows, trying to import it caused
  an error
- issue #81


tarruda:
Thanks

PR #91, opened 2015-02-13
fix #90

### LINKED ISSUE #90: The `reverse` attr actually means `swap`
On [this line](https://github.com/neovim/python-client/blob/master/neovim/ui/gtk_ui.py#L474), the `reverse` attr is interpreted as inverting the rgb color, while I found that it should be `swapping fg and bg` despite of its confusing name:

- [Wikipedia](http://en.wikipedia.org/wiki/ANSI_escape_code) (Search for `inverse or reverse`)
- [Vim](https://github.com/coolwanglu/vim.js/blob/master/src/gui_gtk_x11.c#L3755) 

Not sure if anything should be changed in neovim.

coolwanglu:


tarruda:
Thanks, but please use a more descriptive commit message.

coolwanglu:
Oh, I used backticks, which were not escaped correctly.

coolwanglu:
Updated. Also fixed for the 80-char width checking.

coolwanglu:
I saw errors in Travis CI, but no idea if it's relevant to this PR.

tarruda:
> I saw errors in Travis CI, but no idea if it's relevant to this PR.

These errors are unrelated to this PR, I will fix them with another commit

PR #97, opened 2015-03-25
Provide exception traceback in error response

blueyed:
While NVIM_PYTHON_LOG_FILE can be used to retrieve these, it seems to be
more convenient to know where an error is coming from.

Maybe only `format_exc` should be used.  Please feel free to adopt it as needed.


bfredl:
:+1: LGTM

Later on, I would like some kind debug mode where uncaught exceptions also in async handlers are sent to nvim (with `echoerr` or something as we then have nothing to reply to)

PR #107, opened 2015-04-13
add install_requires 'pyuv'

syswow:
install pyuv by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ from setuptools import setup
 install_requires = [
     'click>=3.0',
     'msgpack-python>=0.4.0',
+    'pyuv',
 ]
 
 if sys.version_info < (3, 4):


tarruda:
AFAIK pyuv is not available on pypy, which is why it's an optional dependency

syswow:
@tarruda Hi, I think we need to install pyuv, since python2 doesn't install asyncio by default.

syswow:
on my Ubuntu machine, when I install asyncio, python2 will throw "SyntaxError: invalid syntax".

tarruda:
Thats the trollius package, which is installed for python < 3.4

syswow:
@tarruda ok, thanks.

tarruda:
@syswow I just uploaded a new version, can you try upgrading it?

syswow:
@tarruda  sure, i have upgraded.

tarruda:
are you still experiencing the error?

syswow:
no error
because I made a mistake running `pip2 install asyncio`.

syswow:
@tarruda  But I have a question on example `limit.py`. 
When I try to `:Cmd`, neovim say:
```
Error detected while processing function remote#define#CommandBootstrap:
line    5:
TypeError('command_handler() takes exactly 3 arguments (2 given)',)
```
Do I need to open a new issue? :worried:

tarruda:
Yes, it's a bug in the python host

tarruda:
@syswow it should be fixed in version 0.0.34

syswow:
@tarruda  comfirmed fixed. thanks.

syswow:
@tarruda One more question:
```viml
@neovim.function('Func')
def function_handler(self, args):
    self._increment_calls()
    self.vim.current.line = (
        'Function: Called %d times, args: %s' % (self.calls, args))
```
What is the default scoping prefix of `Func` in `limit.py`? How can I specify the scoping prefix?

tarruda:
> What is the default scoping prefix of Func in limit.py? How can I specify the scoping prefix?

Since the remote plugin infrastructure is implemented in vimscript(autoload/remote), I don't think this is possible(I could be wrong, cc @ZyX-I). 

There is still much room for improvement in [host.vim](https://github.com/neovim/neovim/blob/master/runtime/autoload/remote/host.vim) and [define.vim](https://github.com/neovim/neovim/blob/master/runtime/autoload/remote/define.vim), if you know vimscript and have ideas then consider sending PRs. One feature I considered implementing is  declarative mappings in the remote plugin, something like this:

```python
@neovim.nnoremap('<c-t>')
def mapping_handler(self, args):
    ...
```

This would require implementing `MappingOnChannel`/`MappingOnHost`

PR #110, opened 2015-04-18
[RFC] api: make remote objects hashable on their identity

bfredl:
Useful for associating data with a buffer purely on the python side.

### DIFF
--- a/neovim/api/common.py
+++ b/neovim/api/common.py
@@ -15,6 +15,10 @@ class Remote(object):
         return (hasattr(other, 'code_data') and
                 other.code_data == self.code_data)
 
+    def __hash__(self):
+        """Return hash based on remote object id."""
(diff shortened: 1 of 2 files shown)


bfredl:
ping @tarruda

tarruda:
:+1: LGTM

bfredl:
merged

bfredl:
If scrutinizer thinks a PR introduces a "major issue" why doesn't it say so until _after_ the PR is merged?

tarruda:
It's supposed to warn in the PR itself

tarruda:
In any case, this is a false positive. Python objects are dynamic and can have attributes added at any time

bfredl:
The "bug" I think is that Remote is not marked as an abstract class/mixin somehow.

PR #111, opened 2015-04-18
[RFC] allow async method calls to nvim

bfredl:
Allows to call neovim rpc methods asynchronously. This is useful to be able to make a lot of calls without requiring 2 context switches per call. For instance for `bufhl` a rich highlighter might make thousands of `buffer_add_highlight` at once for a big file.
Not sure of the api though, perhaps `vim.session.request(..., async=True)` is better?

### DIFF
--- a/test/test_events.py
+++ b/test/test_events.py
@@ -19,15 +19,15 @@ def test_receiving_events():
 @with_setup(setup=cleanup)
 def test_sending_notify():
     # notify after notify
-    vim.session.request('vim_command', "let g:test = 3", async=True)
+    vim.command("let g:test = 3", async=True)
     cmd = 'call rpcnotify(%d, "test-event", g:test)' % vim.channel_id
-    vim.session.request('vim_command', cmd, async=True)
+    vim.command(cmd, async=True)
     event = vim.session.next_message()
     eq(event[1], 'test-event')
     eq(event[2], [3])
 
     # request after notify
-    vim.session.request('vim_command', "let g:data = 'xyz'", async=True)
+    vim.command("let g:data = 'xyz'", async=True)
     eq(vim.eval('g:data'), 'xyz')
 
 
--- a/test/test_events.py
+++ b/test/test_events.py
@@ -16,6 +16,20 @@ def test_receiving_events():
     eq(event[1], 'py!')
     eq(event[2], [vim.current.buffer.number])
 
+@with_setup(setup=cleanup)
+def test_sending_notify():
+    # notify after notify
+    vim.session.request('vim_command', "let g:test = 3", async=True)
+    cmd = 'call rpcnotify(%d, "test-event", g:test)' % vim.channel_id
+    vim.session.request('vim_command', cmd, async=True)
+    event = vim.session.next_message()
+    eq(event[1], 'test-event')
+    eq(event[2], [3])
+
+    # request after notify
+    vim.session.request('vim_command', "let g:data = 'xyz'", async=True)
+    eq(vim.eval('g:data'), 'xyz')
+
 
 @with_setup(setup=cleanup)
 def test_broadcast():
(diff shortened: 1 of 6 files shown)


Shougo:
:+1: 

This feature is needed for deoplete.nvim.

test/test_events.py
```diff
@@ -16,6 +16,20 @@ def test_receiving_events():
     eq(event[1], 'py!')
     eq(event[2], [vim.current.buffer.number])
 
+@with_setup(setup=cleanup)
+def test_sending_notify():
+    # notify after notify
+    vim.session.notify('vim_command', "let g:test = 3")
+    cmd = 'call rpcnotify(%d, "test-event", g:test)' % vim.channel_id
```
Shougo: I think you should use `string.format()` instead of `%` op.
`%` is removed in Python3.

Shougo: Oh, other functions uses `%` like this...

ZyX-I: @Shougo `%` is not removed. Python developers were considering deprecating it and removing some time after, but this decision did not pass through. You may see now that https://docs.python.org/3/library/stdtypes.html#old-string-formatting is talking about string formatting method which has some certain issues (first note), but not about deprecated formatting method.

bfredl: Usually I prefer `"{}".format()` myself, but I wanted to be consistent with the surrounding code. If we want it to change, it is better to discuss that in separate PR.

bfredl:
Note this only exposes low-level "session" interface. We probably want to extend some higher-level functions, at least `command` and `eval` and such. We could either have
```
vim.command("MyCommand", async=True)
vim.eval("AsyncFunction()", async=True)
```
or
```
vim.command_async("MyCommand")
vim.eval_async("AsyncFunction()")
```
I prefer the former (and it would be simpler to implement as they could just pass the async flag to a common `session.call(..., async=...)` function), what do you think @tarruda @shougo ?

test/test_events.py
Shougo: I get it.

Shougo:
> I prefer the former (and it would be simpler to implement as they could just pass the async flag to a common session.call(..., async=...) function), what do you think @tarruda @shougo ?

:+1:

bfredl:
Now one can `vim.command("something", async=True)` and similarly `eval`.  Any more obvious candidates for API functions that should support this?

And, BTW, why can't one `from __future__ import python3`? :P I mean such basics as keyword-only arguments...

tarruda:
> Note this only exposes low-level "session" interface. We probably want to extend some higher-level functions, at least command and eval and such. We could either have

Another option would be to regenerate the API classes/methods when connecting, exposing the async parameter for each generated method. I used to do this on the first versions of this library and switched to a more explicit approach, but maybe it is better to generate the API wrappers after all.

> I prefer the former (and it would be simpler to implement as they could just pass the async flag to a common session.call(..., async=...) function), what do you think @tarruda @shougo ?

:+1:

bfredl:
Yes, a sane option is to autogenerate wrappers per default, and allow to override these manually (if necessary). I suspect many third-party client libs want to autogenerate methods, so they are not out of date any time nvim adds a new method.

bfredl:
Sorry for forgetting this, I think we should merge this as-is, `eval` and `command` is probably the most useful to call async (and with #130 `call` as well) and `async` could be added to all methods when/if autogeneration is implemented.

tarruda:
> Sorry for forgetting this, I think we should merge this as-is, eval and command is probably the most useful to call async (and with #130 call as well) and async could be added to all methods when/if autogeneration is implemented.

:+1:

critiqjo:
:+1:

Version bump?!

bfredl:
I tagged 0.0.38, but @tarruda needs to register it to PyPI.

tarruda:
done

PR #120, opened 2015-05-18
[RFC] Disable cythonizing screen.py for Python 3.

fwalch:
The experimental GUI is only supported for Python 2,
so cythonizing screen.py doesn't make sense on Python 3.


bfredl:
LGTM :+1:

PR #121, opened 2015-05-21
[RFC] Python 3 host: Handle byte strings for specs requests.

fwalch:
@Shougo does this work for you?

Possible alternative:

```diff
diff --git a/neovim/plugin/host.py b/neovim/plugin/host.py
index 2526394..2fb2fcb 100644
--- a/neovim/plugin/host.py
+++ b/neovim/plugin/host.py
@@ -61,6 +61,9 @@ class Host(object):
             warn(msg)
             raise Exception(msg)
 
+        if IS_PYTHON3:
+            args = [arg.decode(self._nvim_encoding) if isinstance(arg, bytes)
+                    else arg for arg in args]
         debug('calling request handler for "%s", args: "%s"', name, args)
         rv = handler(*args)
         debug("request handler for '%s %s' returns: %s", name, args, rv)
@@ -75,6 +78,9 @@ class Host(object):
             warn('no notification handler registered for "%s"', name)
             return
 
+        if IS_PYTHON3:
+            args = [arg.decode(self._nvim_encoding) if isinstance(arg, bytes)
+                    else arg for arg in args]
         debug('calling notification handler for "%s", args: "%s"', name, args)
         handler(*args)
```

### DIFF
--- a/neovim/plugin/host.py
+++ b/neovim/plugin/host.py
@@ -33,7 +33,7 @@ class Host(object):
         self._notification_handlers = {}
         self._request_handlers = {
             'poll': lambda: 'ok',
-            'specs': lambda path: self._specs.get(path, []),
+            'specs': self._on_specs_request,
             'shutdown': self.shutdown
         }
         self._nvim_encoding = nvim.options['encoding']
@@ -155,6 +155,11 @@ class Host(object):
             handlers.append(fn)
         if specs:
             self._specs[plugin_path] = specs
+
+    def _on_specs_request(self, path):
+        if IS_PYTHON3 and isinstance(path, bytes):
+            path = path.decode(self._nvim_encoding)
+        return self._specs.get(path, [])
 
     def _configure_nvim_for(self, obj):
         # Configure a nvim instance for obj(checks encoding configuration)


Shougo:
It works.  Thanks!

bfredl:
I like this idea, but don't we want to do this recursively, i e if `rpcnotify(0,"mymethod", "aa", ["bb", "cc"])`, we also want to decode the strings in the list. There is already support code for this in `neovim/api/common.py` for decoding/encoding the arguments/return for api functions, which could be reused.

fwalch:
@bfredl Something like this? Didn't test it, though (not sure how).

```diff
diff --git a/neovim/plugin/host.py b/neovim/plugin/host.py
index 2526394..1a00594 100644
--- a/neovim/plugin/host.py
+++ b/neovim/plugin/host.py
@@ -7,6 +7,7 @@ import os
 import os.path
 
 from ..api import DecodeHook
+from ..api.common import walk
 from ..compat import IS_PYTHON3, find_module
 
 
@@ -53,8 +54,7 @@ class Host(object):
 
     def _on_request(self, name, args):
         """Handle a msgpack-rpc request."""
-        if IS_PYTHON3 and isinstance(name, bytes):
-            name = name.decode(self._nvim_encoding)
+        name = self._decode_if_bytes(name)
         handler = self._request_handlers.get(name, None)
         if not handler:
             msg = 'no request handler registered for "%s"' % name
@@ -62,21 +62,21 @@ class Host(object):
             raise Exception(msg)
 
         debug('calling request handler for "%s", args: "%s"', name, args)
-        rv = handler(*args)
+        rv = self._decode_if_bytes(handler(walk(self._decode_if_bytes,
+                                                self, args)))
         debug("request handler for '%s %s' returns: %s", name, args, rv)
         return rv
 
     def _on_notification(self, name, args):
         """Handle a msgpack-rpc notification."""
-        if IS_PYTHON3 and isinstance(name, bytes):
-            name = name.decode(self._nvim_encoding)
+        name = self._decode_if_bytes(name)
         handler = self._notification_handlers.get(name, None)
         if not handler:
             warn('no notification handler registered for "%s"', name)
             return
 
         debug('calling notification handler for "%s", args: "%s"', name, args)
-        handler(*args)
+        handler(walk(self._decode_if_bytes, self, args))
 
     def _load(self, plugins):
         for path in plugins:
@@ -156,6 +156,11 @@ class Host(object):
         if specs:
             self._specs[plugin_path] = specs
 
+    def _decode_if_bytes(self, obj):
+        if IS_PYTHON3 and isinstance(obj, bytes):
+            obj = obj.decode(self._nvim_encoding)
+        return obj
+
     def _configure_nvim_for(self, obj):
         # Configure a nvim instance for obj(checks encoding configuration)
         nvim = self.nvim
```

bfredl:
Just looked into the "python3" branch of my `nvim-ipy` rplugin again, and realized that I had to add this
```
        if PY3:
            vim = vim.with_hook(
                neovim.DecodeHook(
                    encoding=vim.options['encoding']))
        self.vim = vim
```
even to get api function decoding right.. Perhaps there should be a flag to `neovim.plugin`, like `@neovim.plugin(decode=True)` to enable decoding both for api functions and rpc handlers. (or even be the default, but not sure of that).

fwalch:
@bfredl I don't think I have time to do anything more complicated than the current PR at the moment (I am not really familiar with the code), so if you want to work on that, feel free :-)

Shougo:
@fwalch I think so, too.

bfredl:
Yes, this PR is a improvement by it self, let's merge it and I will look into decoding convenience later.

PR #116, opened 2015-05-04
[RFC] setup.py: Add optional GUI dependencies.

fwalch:
Resolves #115.

PR #122, opened 2015-05-22
[RFC] setup: Fix pynvim dependencies.

fwalch:
Looks like I messed up the dependencies in #116 :-S

On my system, pycairo doesn't have an egg info, so there's no other way than to just remove the dependency from the setup.py?

Tested with `pip2 install . --user --upgrade && ~/.local/bin/pynvim`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ if not platform.python_implementation() == 'PyPy':
 
 if sys.version_info < (3, 0):
     # Experimental GUI only supported for Python 2.
-    extras_require['GUI'] = ['click>=3.0', 'cairo', 'gobject']
+    extras_require['GUI'] = ['click>=3.0', 'pygobject']
     entry_points['console_scripts'] = ['pynvim=neovim.ui.cli:main [GUI]']
 
 setup(name='neovim',


bfredl:
I seem to can install pycairo via pip using
`pip2 install --allow-external pycairo --allow-unverified pycairo pycairo`
(yes three `pycairo`) but it doesn't reckognize the system installed version, as you said. Perhaps a bug with the arch package?

fwalch:
For me, that command fails with:

```
Collecting pycairo
  pycairo is potentially insecure and unverifiable.
  Downloading http://cairographics.org/releases/pycairo-1.10.0.tar.bz2 (246kB)
    100% |████████████████████████████████| 249kB 3.7MB/s 
    Complete output from command python setup.py egg_info:
      File "<string>", line 18, in <module>
    IOError: [Errno 2] No such file or directory: '/tmp/pip-build-YCslZI/pycairo/setup.py'
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-YCslZI/pycairo
```

fwalch:
I think the problem is with pycairo itself, at least I can't find any `.egg-info` for `cairo` or `pycairo` in `/usr/lib/python2.7/dist-packages` on Ubuntu as well.

bfredl:
Hmm, so it is not a well-formed pypi package. Since pygobject needs it at compile-time to support cairo, I'll guess most systems with python-gobject will have python-cairo anyway.

fwalch:
@bfredl So merge this? FWIW, before #116 there were dependencies on neither gobject nor cairo, so.. it's still a bit of an improvement.

tarruda:
> Can we tag a new release after this? I don't like having the Arch packages at a version where the GUI doesn't work (especially because it's my fault). (cc @tarruda)

:+1: 

One question: does the `extra_require` key enforce these dependencies? They should be optional since the GUI is not required to use the python client.

fwalch:
The dependencies are only enforced once you invoke `pynvim` (because of `[GUI]` at https://github.com/neovim/python-client/pull/122/files#diff-2eeaed663bd0d25b7e608891384b7298R32), I believe.

fwalch:
Is this okay merge?

bfredl:
LGTM  :+1: (sorry, I didn't see you ask before)

fwalch:
@tarruda I created a new release, hope that's okay.. can you upload it to PyPi?

tarruda:
> @tarruda I created a new release, hope that's okay.. can you upload it to PyPi?

Done

PR #129, opened 2015-07-14
handle decoding for rplugin methods the same as for api calls

bfredl:
In a python3 rplugin, per default, return values from api functions like `nvim.eval()` will return unicode strings while args to rpc functions (`@neovim.function` etc)  will be passed as bytes, which is inconsistent and surprising. This PR fixes so that args to rpc functions will be decoded the same way, and this can be overidden by `@neovim.encoding` dectorator on the plugin (or individual method) the same way. No change of defaults for python2, where behaviour already is consistent.

ping @shougo @razvanc87 and others developing python3 plugins.

### DIFF
--- a/neovim/api/common.py
+++ b/neovim/api/common.py
@@ -180,6 +180,13 @@ class DecodeHook(SessionHook):
             return obj.decode(self.encoding, errors=self.encoding_errors)
         return obj
 
+    def walk(self, obj):
+        """Decode bytes found in obj (any msgpack object).
+
+        Uses encoding and policy specified in constructor.
+        """
+        return walk(self._decode_if_bytes, obj, None, None, None)
+
 
 class SessionFilter(object):
 
--- a/neovim/plugin/decorators.py
+++ b/neovim/plugin/decorators.py
@@ -73,7 +73,7 @@ def command(name, nargs=0, complete=None, range=None, count=None, bang=False,
         if eval:
             opts['eval'] = eval
 
-        f.nvim_rpc_spec = {
+        f._nvim_rpc_spec = {
             'type': 'command',
             'name': name,
             'sync': sync,
@@ -98,7 +98,7 @@ def autocmd(name, pattern='*', sync=False, eval=None):
         if eval:
             opts['eval'] = eval
 
-        f.nvim_rpc_spec = {
+        f._nvim_rpc_spec = {
             'type': 'autocmd',
             'name': name,
             'sync': sync,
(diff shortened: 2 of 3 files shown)


razvanc87:
How would you use this `@neovim.encoding` decorator? If you add it after say `@neovim.command` as in:
```python
@neovim.command('blabla', nargs='*')
@neovim.encoding
def blabla_handler...
```
The `@neovim.encoding` decorator gives back a function that takes no arguments, so the above can't work. On the other hand if you do
```python
@neovim.encoding
@neovim.command('blabla', nargs='*')
def blabla_handler...
```
The command won't registere after running `:UpdateRemotePlugins`. The same applies for the entire `plugin` class...

bfredl:
If you use it, you need to call it
```
@neovim.encoding()
@neovim.command(...)
...
```
but ordering before/after `command`/`function` shouldn't matter. But you won't need it on python3, as it is the default. (nvim-ipy sets `DecodeHook` because it _used_ to be needed, I'll remove it as well when updating for this change)

bfredl:
unrelated failure on python 2.6

razvanc87:
Thanks. I also had to add the `@nvim.with_hook(neovim.DecodeHook())` thing. Super, I look forward to this PR.

bfredl:
Ping @tarruda Looks like trollius is broken on python 2.6 on travis. Are you OK with this change otherwise ?

tarruda:
> Ping @tarruda Looks like trollius is broken on python 2.6 on travis. Are you OK with this change otherwise ?

:+1:

bfredl:
As this is a breaking change, I'll tag version after merging this.
No changes needed for `script_host.py` (though it could be simplified, I think), though it breaks deoplete.nvim, so I'll send it a PR.

tarruda:
> As this is a breaking change, I'll tag version after merging this.
No changes needed for script_host.py (though it could be simplified, I think), though it breaks deoplete.nvim, so I'll send it a PR.

:+1: 

After you do I will publish to pypi

bfredl:
Done!

tarruda:
@bfredl you missed updating setup.py with the right version, so I will force-repush 0.0.37

bfredl:
@tarruda :+1:

razvanc87:
Just checked this. I think now the module has another inconsistency. Before I was able to do:
```python
buffer = nvim.current.buffer
buffer[len(buffer)] = None
```
But I get an index error. And `buffer[0]` is allowed. On the other hand:
```python
window = nvim.current.window
window.cursor = len(buffer), 0
```
Works and `window.cursor = 0, 0` gives an out of bounds error.

bfredl:
the first part is intentional at least ( neovim/neovim#2846 )

razvanc87:
Allright, it seems that the rest of the things are unaffected. I can understand the rationale about this slicing / selecting decisions, but I sure hope to have it consistent throughout the objects. Thanks.

Shougo:
@bfredl :+1:

PR #130, opened 2015-07-25
[RFC] expose vim_call_function

bfredl:
This exposes a method call syntax for calling vimscript functions

    nvim.viml.MyVimscriptFunction(args)

Not sure `viml` is the best name. And what happens if we support async calls, would it be `nvim.viml_async.Func(...)` or `nvim.viml.Func(..., async=True)`. ?

### DIFF
--- a/neovim/api/nvim.py
+++ b/neovim/api/nvim.py
@@ -126,9 +126,13 @@ class Nvim(object):
         """Evaluate a vimscript expression."""
         return self._session.request('vim_eval', string, async=async)
 
-    def call(self, name, *args):
+    def call(self, name, *args, **kwargs):
         """Call a vimscript function."""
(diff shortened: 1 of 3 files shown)


Shougo:
It is my desired feature!
Wow!

justinmk:
> Not sure viml is the best name.

Maybe `dispatch` or `dispatcher`?

bfredl:
I also thought of `eval` as +eval is the "official" name for the entire vimscript/viml feature (and is one step less when refactoring old-style calls...) Otherwise `funcs` has a kinda namespacey feeling to it, and is perhaps more transparent when the function is not immediately called ( `vim.funcs.SomeName` )

bfredl:
No more discussion, so let's use `funcs`.

PR #137, opened 2015-09-14
Added embeding example to readme.

traverseda:
An example of embedding neovim as a child process, which was non-obvious to me at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,4 +41,11 @@ bridge](http://vimdoc.sourceforge.net/htmldoc/if_pyth.html#python-vim)):
 [1, 2, 3]
 ```
 
+You can embed neovim into your python application instead of binding to a running neovim instance.
+
+```python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,4 +41,11 @@ bridge](http://vimdoc.sourceforge.net/htmldoc/if_pyth.html#python-vim)):
 [1, 2, 3]
 ```
 
+You can embed neovim into your python application instead of binding to a running neovim instance.
+
+```python
+>>> from neovim import attach
+>>> nvim = neovim.attach('child', argv=["/bin/env", "nvim", "--embed"])
```
justinmk: For parallel form with the above example, leave out the `neovim.`. Simply `attach(...)`.

traverseda:
Fixed

justinmk:
Merged

PR #139, opened 2015-10-08
[RFC] tests: Update tests to neovim changes

bfredl:
Test failures from changes in neovim upstream. the test now documents the correct behaviour as fixed in neovim/neovim#3315

### DIFF
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -100,7 +100,7 @@ def test_vars():
 
 @with_setup(setup=cleanup)
 def test_options():
-    eq(vim.options['listchars'], 'eol:$')
+    eq(vim.options['listchars'], 'tab:> ,trail:-,nbsp:+')
     vim.options['listchars'] = 'tab:xy'
     eq(vim.options['listchars'], 'tab:xy')
 
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -54,27 +54,6 @@ def test_strwidth():
     # 19 * 2 (each japanese character occupies two cells)
     eq(vim.strwidth('neovimのデザインかなりまともなのになってる。'), 44)
 
-
-@with_setup(setup=cleanup)
-def test_list_runtime_paths():
-    # Is this the default runtime path list?
-    homedir = os.path.join(os.environ['HOME'], '.nvim')
-    vimdir = vim.eval('$VIM')
-    dflt_rtp = [
-        homedir,
-        os.path.join(vimdir, 'vimfiles'),
-        vimdir,
-        os.path.join(vimdir, 'vimfiles', 'after')
-    ]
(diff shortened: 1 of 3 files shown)


bfredl:
File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/coverage/annotate.py", line 81
        dest.write(u'  ')
                       ^
    SyntaxError: invalid syntax

what.

bfredl:
`u''` wasn't in python 3.2 apparently, so `coverage.py` decided to drop support for py3.2

bfredl:
We tested `&rtp` as well which broke with XDG. Does really make sense to test this in python-client? Even if relevant _for_ python include path, nothing would stop just having this test in the neovim test suite, as it depends on no python-specific information.

bfredl:
now flake8 is sad
```
neovim/api/buffer.py:13:1: D211  No blank lines allowed before class docstring (found 1)
neovim/api/common.py:4:1: D211  No blank lines allowed before class docstring (found 1)
...
```
but locally I don't get this but instead
```
neovim/plugin/decorators.py:25:5: E731 do not assign a lambda expression, use a def
neovim/plugin/host.py:141:17: E731 do not assign a lambda expression, use a def
```
How is flake even configured? Or is lints just added/removed between flake versions like that?

fwalch:
@bfredl In .travis.yml, there's:

```
pip install -q flake8 flake8-import-order flake8-docstrings pep8-naming;
```

Maybe you don't have `flake8-docstrings` installed locally?

bfredl:
Maybe not. It's just that it seem to be an incredingly fragile way to configure a linter. I would expect the chosen lints to be configured by a config file and not by which modules "just happens" to be installed on every dev machine. (so that one gets a warning if an expected module is missing) I will look more into it tomorrow.

Anyway, do you think the code should be changed or should this error `D211` be supressed?

test/test_vim.py
```diff
@@ -54,27 +54,6 @@ def test_strwidth():
     # 19 * 2 (each japanese character occupies two cells)
     eq(vim.strwidth('neovimのデザインかなりまともなのになってる。'), 44)
 
-
-@with_setup(setup=cleanup)
-def test_list_runtime_paths():
-    # Is this the default runtime path list?
-    homedir = os.path.join(os.environ['HOME'], '.nvim')
```
justinmk: Why not just change this to `.config/nvim` ? Why is the test obsolete?

justinmk:
> should this error D211 be supressed

:+1:

test/test_vim.py
bfredl: Because it is not enough. (And env dependent) `nvim -u NONE` and then `set rtp?` has many more entries.

PR #141, opened 2015-11-09
[RFC] Don't override the encoding of msgpack STRings

bfredl:
This is defined to be `utf-8` per msgpack spec and was already the default in `msgpack.Packer`.
Might fix neovim/neovim#3618

### DIFF
--- a/neovim/api/nvim.py
+++ b/neovim/api/nvim.py
@@ -46,9 +46,6 @@ class Nvim(object):
         session.error_wrapper = lambda e: NvimError(e[1])
         channel_id, metadata = session.request(b'vim_get_api_info')
 
-        encoding = session.request(b'vim_get_option', b'encoding')
-        session._async_session._msgpack_stream.set_packer_encoding(encoding)
-
         if IS_PYTHON3:
             hook = DecodeHook()
             # decode all metadata strings for python3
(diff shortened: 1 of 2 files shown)


justinmk:
Hmm. Can you fix the tests?

bfredl:
#139

justinmk:
D'oh, sorry.

justinmk:
Should be green after rebase.

bfredl:
Thanks, rebased.

bfredl:
I wonder should we bump the version to `0.1.xx` to follow the versioning of main `neovim` repo?

fwalch:
> I wonder should we bump the version to 0.1.xx to follow the versioning of main neovim repo?

(General note) We might have to come up with something here.. if we do any changes that are incompatible with the last release, we can't just tag a new release of the Python client. OTOH, we might need to incorporate some changes for the latest Neovim HEAD version at some point.. example would be neovim/neovim#3405.

bfredl:
True. I think we should release `0.1.xx` for versions that are still compatible with neovim `0.1.x`. When/If we make breaking changes on neovim master, then master really is `0.2-dev` and python-client versions maybe should be `0.2.xx` or `0.2.xx-dev`. I have no clue if pypi supports multiple release "channels" or maybe neovim master followers simply have to follow python-client master as well...

justinmk:
> I wonder should we bump the version to 0.1.xx to follow the versioning of main neovim repo?

Let's go with that for now.

PR #146, opened 2015-12-01
make the drawing_area a self-contained nvim widget

llpamies:
On problem I found with the existing UI is that upon resize it resizes the whole window, and it assumes the size of the "nvim widget" (or drawing area) is the same than the window size. If one wants to embed the drawing area in another application with menu bars, status bars, etc, this approach would not work.

This patch does all the resize work relative to the drawing area and not the whole window.


llpamies:
Not passing travis, and realized that mouse click positions need to be taken care of...

PR #147, opened 2015-12-02
make the drawing_area a self-contained nvim widget, to embbed in other apps

llpamies:
As it is now, the GTK UI assumes that the DrawingArea where nvim is rendered takes the whole window. If you put some other widgets in the windows, the size and position of the DrawingArea doesn't match that of the window and nvim doesn't render properly, and the mouse event positions are incorrect. This pull request aims to solve this problem, paving the way to embed NVIM into more complex GTK GUIs.

I attach two screenshots of what this patch allows:
vbox
and:
hbox

### DIFF
--- a/neovim/ui/gtk_ui.py
+++ b/neovim/ui/gtk_ui.py
@@ -86,21 +86,22 @@ class GtkUI(object):
         """Start the UI event loop."""
         bridge.attach(80, 24, True)
         drawing_area = Gtk.DrawingArea()
+        drawing_area.set_events(drawing_area.get_events() |
+                                Gdk.EventMask.BUTTON_PRESS_MASK |
+                                Gdk.EventMask.BUTTON_RELEASE_MASK |
+                                Gdk.EventMask.POINTER_MOTION_MASK |
+                                Gdk.EventMask.SCROLL_MASK)
         drawing_area.connect('draw', self._gtk_draw)
+        drawing_area.connect('configure-event', self._gtk_configure)
+        drawing_area.connect('key-press-event', self._gtk_key)
+        drawing_area.connect('button-press-event', self._gtk_button_press)
+        drawing_area.connect('button-release-event', self._gtk_button_release)
+        drawing_area.connect('motion-notify-event', self._gtk_motion_notify)
+        drawing_area.connect('scroll-event', self._gtk_scroll)
+        drawing_area.set_can_focus(True)
         window = Gtk.Window()
         window.add(drawing_area)
-        window.set_events(window.get_events() |
-                          Gdk.EventMask.BUTTON_PRESS_MASK |
-                          Gdk.EventMask.BUTTON_RELEASE_MASK |
-                          Gdk.EventMask.POINTER_MOTION_MASK |
-                          Gdk.EventMask.SCROLL_MASK)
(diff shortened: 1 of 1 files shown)


neovim/ui/gtk_ui.py
```diff
@@ -152,7 +152,11 @@ def _nvim_resize(self, columns, rows):
         self._cell_pixel_width = cell_pixel_width
         self._cell_pixel_height = cell_pixel_height
         self._screen = Screen(columns, rows)
-        self._window.resize(pixel_width, pixel_height)
+        prev_rect = self._drawing_area.get_allocation()
```
bfredl: It was some time ago I coded gtk, but to me this looks a bit _ad-hoc_. Won't we just adjust the size of the drawing area and let gtk layouting handle the rest? (especially if we intend the widget to embedded in more complex layouts)

bfredl:
Instead of closing the PR (#146) and opening a new one for an update, feel free to just force push to the same branch. Makes the discussion easier to follow.

neovim/ui/gtk_ui.py
llpamies: I think that the way to deal with it cleanly, is to draw nvim in a custom widget instead of using DrawingArea. With a custom widget, we can properly reply to """get_preferred_width""" and """get_preferred_height""" etc, and the GTK lib will deal with the rest. After it had figure out the geometry, we can go and render.

There is some info here:
https://developer.gnome.org/gtk3/stable/GtkWidget.html

That would require to split gtk_ui and have another gtk_widget, that would be used by gtk_ui.py. Does it sound good?

bfredl: Sounds like a good plan. I thought there was a method to set the preferred the size of a GtkDrawingArea (but not as as strict as `gtk_widget_set_size_request` that permanelty sets the minimum allowed size) but I can't find it again.
If we were to develop this gui futher, I think we eventually should  split the gui out from the python client/host repo, as one might want to fork/experiment with the gui, but still follow the upstream python-client code. Ping @tarruda

llpamies: @bfredl I think that is a good idea to fork the gtkui outside python-client. I can look at this and try to have some simple example.

I was also thinking about the resize problem as it is right now and there is something I do not understand about the **_nvim_resize** callback. It seams to me that it is expecting nvim to proactively decide to resize the screen without it being triggered by the user. Is there any case that it can happen? If that never happens we do not need to listen for this callback.

bfredl: without being triggered by the user _via the client_. You can do `set lines=50` for instance.

llpamies: Sure. Thanks.

tarruda: > If we were to develop this gui futher, I think we eventually should split the gui out from the python client/host repo, as one might want to fork/experiment with the gui, but still follow the upstream python-client code. Ping @tarruda

:+1: 

Ideally we'd have a neovim gtk widget in C so it can be used by any language with gtk bindings

llpamies: The C widget sounds even better.

bfredl:
If you or someone else makes progress on this, please open PR to [neovim/python-gui](https://github.com/neovim/python-gui).

PR #148, opened 2015-12-03
convert byte strings in python3 to fix #145

### LINKED ISSUE #145: redraw commands in python3 are buffers not strings
This is a follow-up of #144. What happened is that in python3 the redraw instructions from nvim session are returned as byte buffers and not strings, and all the parsing expects strings. Besides that the GUI seems to work.

Is it supposed to work with python3? is anyone looking to fix this? I can take a look at it...

llpamies:
This is a simple fix to make the CLI work with python3. I took the code from one of the tests. It fixes #145.

### DIFF
--- a/neovim/ui/cli.py
+++ b/neovim/ui/cli.py
@@ -5,7 +5,8 @@ import click
 
 from .ui_bridge import UIBridge
 from .. import attach
-
+from ..api import DecodeHook
+from ..compat import IS_PYTHON3
 
 @click.command(context_settings=dict(allow_extra_args=True))
 @click.option('--prog')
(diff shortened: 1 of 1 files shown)


neovim/ui/cli.py
```diff
@@ -58,6 +60,11 @@ def main(ctx, prog, notify, listen, connect, profile):
         nvim_argv = shlex.split(prog or 'nvim --embed') + ctx.args
         nvim = attach('child', argv=nvim_argv)
 
+    if IS_PYTHON3:
+        # For Python3 we decode binary strings as Unicode for compatibility
+        # with Python2
```
bfredl: doesn't really make sense, we are doing this for compability with python3 :)

bfredl:
Sorry for late reply. Tested this now and it works fine, so merged, execpt that I removed one comment as wasn't really correct and shouldn't be needed anyway. Thanks!

justinmk:
@bfredl @fwalch @tarruda Was this published to pypi?

bfredl:
@justinmk I'm tagging a release.

bfredl:
I tagged but @tarruda needs to push to PyPI.

PR #153, opened 2016-01-03
Do not crash applicatoin due to un-implemented features

harish2704:
ignore those events wih proper warning message

### DIFF
--- a/neovim/ui/gtk_ui.py
+++ b/neovim/ui/gtk_ui.py
@@ -125,6 +125,7 @@ class GtkUI(object):
         GObject.idle_add(wrapper)
 
     def _nvim_update_menu( self, *args ):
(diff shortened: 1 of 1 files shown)


neovim/ui/gtk_ui.py
```diff
@@ -124,6 +124,11 @@ def wrapper():
             self._screen_invalid()
         GObject.idle_add(wrapper)
 
+    def _nvim_update_menu( self, *args ):
```
bfredl: this should be placed among the other `_nvim_` event handlers. Also just `pass` is fine here, compare `_nvim_visual_bell` (we won't expect all guis to respond to all ui updates)

bfredl:
superseded by neovim/python-gui#2

PR #154, opened 2016-01-04
`logging.warn()` is deprecated and removed in Py3.5

s0undt3ch:


bfredl:
Merged, thanks. It is not removed yet, though, so I updated the title.

s0undt3ch:
It is on 3.5.1, at least, I got a traceback because of it...

bfredl:
Hmm, it isn't on 3.5.0 (I tested) and is not mentioned in 3.5.1 relnotes... Anyway this change was obviously correct as `warning` is preferred even back 2.6 (oldest version we support)

s0undt3ch:
Woa!

```python
Python 3.4.3 (default, May  3 2015, 19:19:52) 
[GCC 4.9.2 20150304 (prerelease)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> logging.warn
<function warn at 0x7f0da2f43400>
>>> 
```
```python
Python 3.5.1 (default, Dec  7 2015, 12:58:09) 
[GCC 5.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> logging.warn
<function warn at 0x7fbe07e7ebf8>
>>> 
```

Now, why did I get that traceback and why did it stop happening after this change!? Weird....

s0undt3ch:
I believe the issue was probably triggered by trying to start the GtkUI from python3(when I now known that it currently only supports python 2)

bfredl:
It is supposed to work on python3 as well. could you show complete traceback?

PR #160, opened 2016-02-17
typos in the gtk key table

timeyyy:


bfredl:
Strange, these keys work fine even with the errors in the table. Anyway, thanks for the fix.

Issue #152, opened 2015-12-18
Python client crashes under windows


RedX2501:
Reported under neovim/neovim/issues/3862 and equalsraf/neovim-qt/issues/79

When starting the latest build from equalsraf repo python will not start properly.

The command executed was `:py 42`

My question is how do you debug this? Python quits to fast for me to able able to attach...

The python log file contains:

```
2015-12-16 19:30:08,586 [ERROR @ base_events.py:default_exception_handler:1053] 16700 - Fatal read error on pipe transport
protocol: <neovim.msgpack_rpc.event_loop.asyncio.AsyncioEventLoop object at 0x027314B0>
transport: <_ProactorReadPipeTransport fd=0>
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\trollius\proactor_events.py", line 195, in _loop_reading
    self._read_fut = self._loop._proactor.recv(self._sock, 4096)
  File "C:\Python27\lib\site-packages\trollius\windows_events.py", line 428, in recv
    self._register_with_iocp(conn)
  File "C:\Python27\lib\site-packages\trollius\windows_events.py", line 609, in _register_with_iocp
    _overlapped.CreateIoCompletionPort(obj.fileno(), self._iocp, 0, 0)
WindowsError: [Error 6] Das Handle ist ungültig
```

The neovim log file nvimlog:

```
015/12/16 20:43:55 [error @ call_set_error:743] 17716 - msgpack-rpc: Channel was closed by the client
2015/12/16 20:44:33 [info @ main:533] 16324 - Starting Neovim main loop.
2015/12/16 20:44:41 [error @ forward_stderr:333] 16324 - Channel 2 stderr: Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Python27\lib\site-packages\neovim\__init__.py", line 64, in start_host
    session = stdio_session()
  File "C:\Python27\lib\site-packages\neovim\msgpack_rpc\__init__
2015/12/16 20:44:42 [error @ forward_stderr:333] 16324 - Channel 2 stderr: .py", line 36, in stdio_session
    return session('stdio')
  File "C:\Python27\lib\site-packages\neovim\msgpack_rpc\__init__.py", line 17, in session
    loop = EventLoop(transport_type, *args, **kwargs)
  File "C:\Python27\lib\site-packages\neovim\m
[...]

PR #167, opened 2016-02-23
Don't use BaseException.message @equalsraf #152

maxwell-k:
This patch helps show the underlying error in [#152](https://github.com/neovim/python-client/issues/152)

BaseException.message was only available in Python 2.5:
https://www.python.org/dev/peps/pep-0352/#retracted-ideas

This was patch was proposed in neovim/neovim #4253 by @equalsraf :
https://github.com/neovim/neovim/issues/4253#issuecomment-184437483

Without the change calling BaseException.message hides any errors behind an AttributeError:
AttributeError: 'OSError' object has no attribute 'message'

### DIFF
--- a/neovim/msgpack_rpc/event_loop/asyncio.py
+++ b/neovim/msgpack_rpc/event_loop/asyncio.py
@@ -44,7 +44,7 @@ class AsyncioEventLoop(BaseEventLoop, asyncio.Protocol,
 
     def connection_lost(self, exc):
         """Used to signal `asyncio.Protocol` of a lost connection."""
-        self._on_error(exc.message if exc else 'EOF')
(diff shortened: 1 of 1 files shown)


bfredl:
This works in both python 2 and 3 right?

maxwell-k:
That's absolutely right, I have tested on 64 bit Windows 7 with Python 3.4.3 on 2.7.11.

Both `BaseException.message` and `BaseException.args` are available on Python 2.7 but only `BaseException.args` is available on Python 3. Some background is available on [python-dev](https://mail.python.org/pipermail/python-dev/2007-April/072542.html).

bfredl:
Ok, makes sense, thanks.

PR #169, opened 2016-02-24
better handling of exceptions in async contexts

bfredl:
This oneliner crashes the python client:
`py vim.session.threadsafe_call(lambda: [vim.eval("3"),undefined_variable])`
yes the `vim.eval("3")` is needed.

Print exceptions arising from async plugin handler, as already is done for sync ones.

Define `nvim.threadsafe_call` a wrapper around `nvim.session.threadsafe_call`) that prints uncaught exceptions in the async call, as well as the context the call came from. Also I don't think we should recommend plugins to call methods on `nvim.session` if they can avoid it, it's better to put all exported methods, including `threadsafe_call` on the api objects.

Docs are missing, so WIP.

### DIFF
--- a/neovim/__init__.py
+++ b/neovim/__init__.py
@@ -16,7 +16,7 @@ from .plugin import (Host, autocmd, command, encoding, function, plugin,
 __all__ = ('tcp_session', 'socket_session', 'stdio_session', 'child_session',
            'start_host', 'autocmd', 'command', 'encoding', 'function',
            'plugin', 'rpc_export', 'Host', 'DecodeHook', 'Nvim',
-           'SessionHook', 'shutdown_hook', 'attach')
+           'SessionHook', 'shutdown_hook', 'attach', 'setup_logging')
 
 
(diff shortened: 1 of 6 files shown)


bfredl:
I wonder if `threadsafe_call` should be `async_call` because it also makes sense to call it in a sync request handler in the same thread like this
```
    @neovim.function("IPyConnect", sync=True)
    def ipy_connect(self, args):
        # do stuff that would be racy if it was an async contexts
        # (configuring an output buffer in my case)
        self.do_sync_stuff(args)
        # do the rest asynchronously 
        # (waiting on an external process, so should be async)
        self.vim.async_call(self.do_async_stuff,args) 
```
or even in an nvim-aware old-style plugin
```
if has('nvim')
   :python vim.async_call(async_function)
else
   :python sync_function()
end if
```
only problem it might be confusing with `nvim.call(..., async=True)`

mhinz:
:+1:

PR #172, opened 2016-02-26
Rename common.py to test_common.py (and fix imports)

### LINKED ISSUE #170: .tar.gz files uploaded to PyPI is missing test/common.py
The 0.1.2 tar.gz is missing test/common.py, and the tests can't be run.

(Maybe it is silly to say, but If you want to republish the tar.gz, please
bump the version instead of changing the file, to not mess with automatic tools
that harvest tar.gzs).

viccuad:
This makes the file be included on tarballs by the build system.

Fixes #170


bfredl:
Thanks!

PR #173, opened 2016-02-27
move the gui to a separate repo

bfredl:
This seems better than messing with "optional" features and dependencies.
The python-client should focus on being a client library and the provider for python plugins.

a preview of the separate gui repo can be found at https://github.com/neovim/python-gui . It should not be considered "master" until this PR is merged. 

Contrary to what I said before, I think it is better to keep the ui_bridge.py along with the gui, as it is not really general but for one specific strategy of integrating the event loops. Other gui implementations can just copy it if they want to reuse it.


tarruda:
[Done](https://github.com/neovim/python-gui)

> The python-client should focus on being a client library and the provider for python plugins.

:+1: 

Ideally the python host should also split into another repository. It adds dependencies like greenlet that are not necessary for simple, non-plugin clients.

bfredl:
> Ideally the python host should also split into another repository. It adds dependencies like greenlet that are not necessary for simple, non-plugin clients.

Personally I think that is going too far. I suppose almost all users of nvim that would use the python-client in some way, would also expect (legacy or new style) python plugins to work. OTOH quite few that use python plugins would actually use this gui (it doesn't bring that much over a truecolor terminal, but it is nice to have as a simple test gui)

bfredl:
Could you turn on travis as well (we don't have tests yet, but lets use flake and the scrutinizer) ?

tarruda:
> Could you turn on travis as well (we don't have tests yet, but lets use flake and the scrutinizer) ?

:+1: 

Strangely enough, this repository seem to be deactivated(even though tests are being ran):

travis

bfredl:
@tarruda could you change the default branch of `python-gui` to `master` ? It would be good to activate the scrutinizer as well.

tarruda:
:+1:

Issue #175, opened 2016-03-02
README: a tad more documentation please


mhinz:
Could you please add one or two sentences to each of the following points? Would probably help newbies or people that usually don't use Python. :)
1. How to run the tests?
2. How to make Neovim use `master` instead of whatever `pip` installed?

(Basically how to set up a proper environment for working at the Python client.)

bfredl:
:+1: we should also have an overview of the new features relative vim's if_python and eventually set up sphinx docs or something

bfredl:
Some improvements merged, but still some missing pieces like logging/troubleshooting.

mhinz:
I like what I see so far! :+1:

justinmk:
I guess https://github.com/neovim/python-client/commit/2db41a5a881fa899ae0108a0ba0ca7dd15bb2a67 covers this.

PR #177, opened 2016-03-05
[WIP] API/dev documentation in README.md

bfredl:
Adressing #175.

I was going to suggest to `export PYTHONPATH=/path/to/python-client` to try master version without installing, but packages installed `--user` apparently comes before `$PYTHONPATH` which makes the variable useless. Anyone knows the quickest way to prepend a folder to `sys.path` from an env var?

The bullet list will do for now, but will not scale obviously. We should set up sphinx or similar for API documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,80 @@ Library for scripting Nvim processes through its msgpack-rpc API.
 pip install neovim
 ```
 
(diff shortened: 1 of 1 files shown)


bfredl:
By no means done but let's merge this and improve iteratively.

PR #179, opened 2016-03-06
Refactor: let Nvim be the session wrapper and simplify API remote 

bfredl:
Refactor the API implemention to not use SessionFilter. I don't see any point of sticking in filters "between" `Session` and `Nvim`,  `Nvim` could simply be the wrapper around session, and we don't need the Session object to be arbitrarily configurable. Legacy `vim.eval` behavior is better expressed as a subclass, so that `vim.api.eval` always can give the new, consistent behavior.

Also some experiments of using descriptors to make API wrapping less verbose. Though we need a solution for documentation, I wonder if one could just set `__doc__` on a descriptor object (not the descriptor class) and have it work.


bfredl:
removed the descrptor tests for now to keep it somewhat focused.

tarruda:
:+1:

bfredl:
With this change `nvim.session` is unexported. Upgrade as follows:
`nvim.session.next_message` -> `nvim.next_message`
`nvim.session.run` -> `nvim.run_loop`
`nvim.session.stop` -> `nvim.stop_loop`
`nvim.session.request` -> `nvim.request` (or `nvim.api.method` )

PR #180, opened 2016-03-08
Script host

bfredl:
I just realized we can mitigate the version incompatibility issue. This version of the python-client is compatible with neovim both before and after neovim/neovim#4418. So what we should do is:

* merge this and release python-client 0.1.4
* wait a bit
* merge neovim/neovim#4418
* wait a bit
* merge the rest of #179


mhinz:
:+1:

justinmk:
:+1:

PR #182, opened 2016-03-09
[RFC] README: use tw=80 for text

mhinz:


mhinz:
I thought I could push it myself, but apparently I have to be explicitly added as contributor? Anyway, PRs work just as well. :)

bfredl:
Thanks. I think the write access can be set per repo.

PR #186, opened 2016-03-18
Range: The end of the range should point at the last line of the range

### LINKED ISSUE #185: Vim inconsistency: Range object's end attribute points at the line after the buffer
To verify, compare running (both in Vim and Neovim) the following commands:

:python import vim; print vim.current.range.start, vim.current.range.end

In Vim: 0 0
In Neovim: 0 1

The behaviour is the same when the visual range is actually selected:

In Vim: 0 4
In Neovim: 0 5

tbabej:
To preserve compatibility with Vim, the end of the range should point
at the last line of the range, not at the first line below.

The inconsistency can be demostrated by running (both in Vim and Neovim):

:python import vim; print vim.current.range.start, vim.current.range.end

either with visual range selected or without it.

Fixes #185.

### DIFF
--- a/neovim/api/buffer.py
+++ b/neovim/api/buffer.py
@@ -166,10 +166,10 @@ class Range(object):
     def __init__(self, buffer, start, end):
         self._buffer = buffer
         self.start = start - 1
-        self.end = end
+        self.end = end - 1
 
     def __len__(self):
-        return self.end - self.start
+        return self.end - self.start + 1
 
     def __getitem__(self, idx):
         if not isinstance(idx, slice):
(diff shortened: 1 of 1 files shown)


bfredl:
Thanks! Not very pythonic, but as this api is mainly for vim compatibility anyway...

tbabej:
@bfredl I agree. One option would be to introduce a Range.end property, and having the other methods use Range._end internal attribute instead. I just didn't pick that option since it seemed as an overkill for this particular problem.

bfredl:
Yes, this solution is fine, no need to overdo it.

PR #187, opened 2016-03-19
host: more robust way of finding the script host

bfredl:
Some people seem to have trouble with host finding the script host. Let's just use ordinary imports for this.

### DIFF
--- a/neovim/plugin/script_host.py
+++ b/neovim/plugin/script_host.py
@@ -5,7 +5,8 @@ import logging
 import os
 import sys
 
-import neovim
+from .decorators import plugin, rpc_export
+from ..api import SessionHook
 
 __all__ = ('ScriptHost',)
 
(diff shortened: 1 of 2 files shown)


neovim/plugin/script_host.py
```diff
@@ -59,9 +60,6 @@ def setup(self, nvim):
 
     def teardown(self):
         """Restore state modified from the `setup` call."""
-        for plugin in self.installed_plugins:
```
bfredl: @tarruda This is leftover code that is safe to remove right?

tarruda: Actually, it was there for supporting teardown hooks so plugins had a chance to cleanup or save state when the host is closes, but I guess it was unused so it should be fine to remove.

bfredl: but surely that should have been in `host.py` and not in `script_host.py` ?

Issue #184, opened 2016-03-18
vim.funcs.getchar() chokes on special input on python3


fmoralesc:
Compare
``` vim
py print vim.funcs.getchar()
```
and
``` vim
py3 print(vim.funcs.getchar())
````
when you press backspace (for example).

I get
```
Se ha detectado un error al procesar function provider#python3#Call:                                                              
línea   18                                                                                                                        
UnicodeDecodeError('utf-8', b'\x80kb', 0, 1, 'invalid start byte')                                                                
Traceback (most recent call last):                                                                                                
  File "/usr/lib/python3.5/site-packages/neovim/msgpack_rpc/session.py", line 181, in handler                                     
    rv = self._request_cb(name, args)                                                                                             
  File "/usr/lib/python3.5/site-packages/neovim/api/common.py", line 237, in filter_request_cb                                    
    walk(self._in, args, self, name, 'request'))                                                                                  
  File "/usr/lib/python3.5/site-packages/neovim/plugin/host.py", line 66, in _on_request                                          
    rv = handler(*args)                                                                                                           
  File "/usr/lib/python3.5/site-packages/neovim/plugin/host.py", line 152, in <lambda>                                            
    decoder = lambda fn, hook, *args: fn(*hook.walk(args))                                                                        
  File "/usr/lib/python3.5/site-packages/neovim/plugin/script_host.py", line 77, in python_execute                                
    exec(script, self.module.__dict__)                                                                                            
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte                 
```

bfredl:
we can add `decode=False` kwarg, should be simple with #179

bfredl:
should be possible to get bytes input on 0.1.6 using `vim.funcs.getchar(decode=False)`

fmoralesc:
Cool. Thanks for the heads up.

PR #189, opened 2016-03-24
remove decodehook

bfredl:
While doing breaking changes ( #179 ) ...

Per decision, rplugins do not need support `encoding != utf-8`. (But it should not be too hard to add back if we really want this.) But setting the encoding (and not just an on/off switch) per rplugin definitely makes no sense.

Allows to turn on/off decoding per request, ref #184 .

### DIFF
--- a/neovim/__init__.py
+++ b/neovim/__init__.py
@@ -6,16 +6,16 @@ import logging
 import os
 import sys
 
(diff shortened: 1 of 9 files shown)


justinmk:
:+1:

PR #193, opened 2016-04-02
nvim.session.threadsafe_call is quite common. Support it a bit longer.

bfredl:


justinmk:
In https://gist.github.com/tarruda/37f7a3e22996addf8921#gistcomment-1756929 the `run()` method is used. Can that be added to `CompatibilitySession` ?

Actually, getting that gist to work as-is would be a good test of compatibility.

PR #197, opened 2016-04-16
Fix `bang` and `register` command modifier

zauguin:
Defining a command with `bang=True` leads to a .init.vim-rplugin~ file which Neovim can't read.
I'm not sure if this belongs here or if it is a Neovim bug, but it's quite easy to fix.


bfredl:
Looks consistent with how we handle `range=True'. Thanks!

PR #200, opened 2016-04-30
update README

dmerejkowsky:
Added some docs about `subscribe`.

Note that right now I use a `while True` loop, I'm not sure if there's a better way

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,22 @@ below.
   wait for nvim to complete the request, which also means that the return value
   is unavailable.
 
+### Subscribe to events
+
+You can subscribe to Neovim events (different from Vim's autocmd events)
+by using subscribe:
+
+```python
+vim.subscribe("my-event")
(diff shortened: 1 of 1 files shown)


bfredl:
`next_message()` might work for a one-shot script but the recommended way is:
```
 def on_setup():
    pass

def on_request(method, args):
    pass

def on_notification(method, args):
    pass

nvim.run_loop(on_request, on_notification, on_setup)
```
the point of `on_setup` is that the code will run inside the event loop, so a blocking request to nvim inside it can invoke a nested request back to the same client.

This should probably be below the mention of `attach` to be clear that this is useful from an external client and not an rplugin. (Update the PR if you like, otherwise I will fix the docs).

dmerejkowsky:
> Update the PR if you like, otherwise I will fix the docs

Hum. Not sure I understand everything, so please do :)

PR #201, opened 2016-05-05
Fix failing tests on Windows.

brcolow:
Test "test_vim.test_command" was failing because of the call to
os.unlink(..), as the file was still open in vim. Adding a conditional
to execute the call only if we are not running under Windows was the
most straight-forward "fix".

Test "test_vim.test_chdir" was failing for two reasons. Firstly because
of missing Windows-specific handling of path names in
find_file_in_path_option in file_search.c in Neovim core. Secondly
because although one can indeed `:cd /` on Windows (or at least, should
be able to), a corresponding call to `:pwd` will echo the root drive
of the current working directory of Neovim.

test_chdir should work now that neovim/neovim#4711 was merged.

### DIFF
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -21,7 +21,8 @@ def test_command():
     vim.command('normal itesting\npython\napi')
     vim.command('w')
     ok(os.path.isfile(fname))
-    eq(open(fname).read(), 'testing\npython\napi\n')
+    with open(fname) as f:
+      eq(f.read(), 'testing\npython\napi\n')
     os.unlink(fname)
 
 
@@ -63,8 +64,10 @@ def test_strwidth():
 @with_setup(setup=cleanup)
 def test_chdir():
     pwd = vim.eval('getcwd()')
+    root = os.path.abspath(os.sep)
+    # We can chdir to '/' on Windows, but then the pwd will be the root drive
     vim.chdir('/')
-    eq(vim.eval('getcwd()'), '/')
+    eq(vim.eval('getcwd()'), root)
     vim.chdir(pwd)
     eq(vim.eval('getcwd()'), pwd)
 


test/test_vim.py
```diff
@@ -22,7 +22,8 @@ def test_command():
     vim.command('w')
     ok(os.path.isfile(fname))
     eq(open(fname).read(), 'testing\npython\napi\n')
-    os.unlink(fname)
+    if os.name != 'nt':
```
fwalch: > the file was still open in vim

What about adding `vim.command('q')` here, then?

brcolow: Tried it before posting the issue actually, didn't fix it =(

test/test_vim.py
```diff
@@ -63,8 +64,9 @@ def test_strwidth():
 @with_setup(setup=cleanup)
 def test_chdir():
     pwd = vim.eval('getcwd()')
-    vim.chdir('/')
```
justinmk: Vim/Neovim should work just fine with `/` as the root path. I just tried `:cd /` in Windows gvim, it cd'd to `C:\`.

test/test_vim.py
```diff
@@ -22,7 +22,8 @@ def test_command():
     vim.command('w')
     ok(os.path.isfile(fname))
     eq(open(fname).read(), 'testing\npython\napi\n')
-    os.unlink(fname)
+    if os.name != 'nt':
+        os.unlink(fname)
```
justinmk: Why shouldn't this work in Windows?

brcolow: It throws a Win32 API error saying that the temporary file is still in use.

test/test_vim.py
brcolow: Ahhh..yes...you can change the directory using '/' as the root path, which will then change the directory to the root drive, but then `:pwd` will print the drive, not `/`. I can edit the test to reflect this and make it more clear.

justinmk: Appears to be a problem in the core. I have an old version of neovim-qt (`0.1.3-dev`), and `:cd /` gives `E472: Command failed`. We're probably missing some Windows-specific handling somewhere.

brcolow: Thanks to @equalsraf and @jamessan we found the missing bit of Windows-specific handling.

From vim:
ttps://github.com/vim/vim/blob/0b9e4d1224522791c0dbbd45742cbd688be823f3/src/misc2.c#L5445

Which should belong in Neovim's:
https://github.com/neovim/neovim/blob/master/src/nvim/file_search.c#L1404 

Pending commit to Neovim's #810.

bfredl:
is this ready to merge?

brcolow:
I think so...the question from @justinmk still remains about why the unlink doesn't work on Windows, and the suggestion by @fwalch about using `vim.command('q')` (which doesn't work)...in other words: this makes the tests pass, but I don't know if removing the `unlink` line will have any unintended consequences. The other test though (test_vim.test_chdir) is :+1:. So just need a confirmation/decision about test_vim.test_command about removing the `unlink` call.

test/test_vim.py
equalsraf: in the previous suggestion was for sending 'q', maybe that call to unlink is made before neovim actually shuts down. just for testing purposes try sleeping for a second before calling unlink just to be sure
-

brcolow: No amount of sleeping seems to have any affect..still results in error:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\brcolow\\AppData\\Local\\Temp\\tmp9zm0bm83'`

test/test_vim.py
```diff
@@ -22,7 +22,8 @@ def test_command():
     vim.command('w')
     ok(os.path.isfile(fname))
     eq(open(fname).read(), 'testing\npython\napi\n')
```
justinmk: @brcolow Try changing this so that the `open()`'d file gets closed before we try `os.unlink()` below. http://askubuntu.com/a/701536 Perhaps this:

```python
with open(fname) as f:
    eq(f.read(), 'testing\npython\napi\n')
```

Then we shouldn't need to check `os.name != 'nt'` below.

test/test_vim.py
```diff
@@ -21,8 +21,8 @@ def test_command():
     vim.command('normal itesting\npython\napi')
     vim.command('w')
     ok(os.path.isfile(fname))
-    eq(open(fname).read(), 'testing\npython\napi\n')
-    os.unlink(fname)
```
justinmk: this line shouldn't be removed, otherwise there's no reason to make the change

test/test_vim.py
```diff
@@ -63,8 +63,10 @@ def test_strwidth():
 @with_setup(setup=cleanup)
 def test_chdir():
     pwd = vim.eval('getcwd()')
+    root = os.path.abspath(os.sep)
```
justinmk: shouldn't this call be made _after_ the call to `vim.chdir('/')` below?

PR #203, opened 2016-05-08
[WIP] Use pyuv on Windows

brcolow:
Very, very, very WIP.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 import platform
 import sys
+import os
 
 from setuptools import setup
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,6 +16,9 @@
     # pypy already includes an implementation of the greenlet module
     install_requires.append('greenlet')
 
+if os.name == 'nt':
```
bfredl: move this before line 10 and change it to `elif sys.version_info < (3, 4):` (trollius is not required if we have pyuv)

brcolow: Done

PR #209, opened 2016-06-06
vim.buffers is a map not a list.

asankah:
`python-buffers` describes `vim.buffers` as a mapping object. `python-buffer`
documents that `b.number` where `b` is a Buffer object can be used as a
`python-buffers` key. I.e. `b == vim.buffers[b.number]` should be true if `b`
is a valid buffer object.

`nvim.py` was implementing `vim.buffers` as a `RemoteSequence` which doesn't
provide the aforementioned properties. This change implements it as a
`RemoteSequenceAsMap` object instead. The latter is a new class that
allows a sequence returned via a single msgpack-rpc call to be exposed
to Python as a map.

### DIFF
--- a/neovim/api/nvim.py
+++ b/neovim/api/nvim.py
@@ -79,7 +79,7 @@ class Nvim(object):
         self.vars = RemoteMap(self, 'vim_get_var', 'vim_set_var')
         self.vvars = RemoteMap(self, 'vim_get_vvar', None)
         self.options = RemoteMap(self, 'vim_get_option', 'vim_set_option')
-        self.buffers = RemoteSequence(self, 'vim_get_buffers')
+        self.buffers = Buffers(self)
         self.windows = RemoteSequence(self, 'vim_get_windows')
         self.tabpages = RemoteSequence(self, 'vim_get_tabpages')
         self.current = Current(self)
@@ -329,6 +329,41 @@ class Nvim(object):
                 self._err_cb(msg)
                 raise
         self._session.threadsafe_call(handler)
+
+
+class Buffers(object):
+
+    """Remote NVim buffers.
+
+    Currently the interface for interacting with remote NVim buffers is the
+    `vim_get_buffers` msgpack-rpc function. Most methods fetch the list of
+    buffers from NVim.
+
+    Conforms to *python-buffers*.
+    """
+
+    def __init__(self, nvim):
+        """Initialize a Buffers object with Nvim object `nvim`."""
+        self._fetch_buffers = nvim.api.get_buffers
+
+    def __len__(self):
+        """Return the count of buffers."""
+        return len(self._fetch_buffers())
+
+    def __getitem__(self, number):
+        """Return the Buffer object matching buffer number `number`."""
+        for b in self._fetch_buffers():
+            if b.number == number:
+                return b
+        raise KeyError(number)
+
+    def __contains__(self, b):
+        """Return whether Buffer `b` is a known valid buffer."""
+        return isinstance(b, Buffer) and b.valid
+
+    def __iter__(self):
+        """Return an iterator over the list of buffers."""
+        return iter(self._fetch_buffers())
 
 
 class CompatibilitySession(object):
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -92,13 +92,29 @@ def test_options():
 
 @with_setup(setup=cleanup)
 def test_buffers():
+    buffers = []
+
+    # Number of elements
     eq(len(vim.buffers), 1)
-    eq(vim.buffers[0], vim.current.buffer)
+
+    # Indexing (by buffer number)
+    eq(vim.buffers[vim.current.buffer.number], vim.current.buffer)
+
+    buffers.append(vim.current.buffer)
     vim.command('new')
     eq(len(vim.buffers), 2)
-    eq(vim.buffers[1], vim.current.buffer)
-    vim.current.buffer = vim.buffers[0]
-    eq(vim.buffers[0], vim.current.buffer)
+    buffers.append(vim.current.buffer)
+    eq(vim.buffers[vim.current.buffer.number], vim.current.buffer)
+    vim.current.buffer = buffers[0]
+    eq(vim.buffers[vim.current.buffer.number], buffers[0])
+
+    # Membership test
+    ok(buffers[0] in vim.buffers)
+    ok(buffers[1] in vim.buffers)
+    ok({} not in vim.buffers)
+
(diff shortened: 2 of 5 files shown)


bfredl:
Actually it is some weird hybrid, getitem is indexing by buffer ids, but iterating should still yield the actual buffer objects. Also `items()`/`keys()` etc does not exist. Your `__getitem__` implemention should work, but more efficient (less roundtrips) would be to add a `vim_get_buffer_by_id` method.

asankah:
Indeed. I'll update the patch to conform to python-buffers. I guess we can't make it a real mapping object.

What's the protocol for adding a vim_get_buffer_by_id method? I can add it to neovim. But once the corresponding change is committed to python-client, then the neovim python package would only work with the latest neovim containing the vim_get_buffer_by_id implementation. Is there a graceful way to handle the transition or is it expected that folks keep their neovim build up-to-date?

bfredl:
We can merge the "slow" version  into python-client right now.

we can merge `vim_get_buffer_by_id` to neovim master, but only once the next neovim version is released,  will we will release a version of the python-client that's using it. It is understood that the latest released nvim version might require the latest python-client version or vice-versa (at some point we should add a version check) but if we can make the incompatibility "one-way" we should do it.

asankah:
Have another look?

neovim/api/buffers.py
```diff
@@ -0,0 +1,53 @@
...
+The key for indexing into vim.buffers[] is the buffer number as described in
+*python-buffer*.
+"""
+
+import functools
+
+__all__ = ('Buffers')
+
+
+class Buffers(object):
```
bfredl: should be in `api/nvim.py` as it is tightly coupled to the nvim class. compare `nvim.current`

neovim/api/buffers.py
```diff
@@ -0,0 +1,53 @@
...
+    def __getitem__(self, number):
+        """Return the Buffer object matching buffer number `number`."""
+        for b in self._fetch_buffers():
+            if b.number == number:
+                return b
+        raise KeyError("No buffers match buffer number {}".format(number))
+
+    def __contains__(self, b):
+        """Return whether Buffer `b` is a known valid buffer."""
+        return b in self._fetch_buffers()
```
bfredl: to avoid fetching the buffers we could do `isinstance(b, Buffer) and b.valid`

neovim/api/buffers.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __iter__(self):
+        """Return an iterator over the list of buffers."""
+        return self._fetch_buffers().__iter__()
```
bfredl: `return iter(self._fetch_buffers())`

test/test_vim.py
```diff
@@ -92,13 +92,33 @@ def test_options():
...
+    vim.current.buffer = buffers[0]
+    eq(vim.buffers[vim.current.buffer.number], buffers[0])
+
+    # Membership test
+    ok(buffers[0] in vim.buffers)
+    ok(buffers[1] in vim.buffers)
+    ok({} not in vim.buffers)
+
+    # Iteration
+    eq(buffers, [b for b in vim.buffers])
```
bfredl: `eq(buffers, list(vim.buffers))` you can remove the below as it is equivalent.

neovim/api/buffers.py
asankah: Done.

neovim/api/buffers.py
asankah: Done.

neovim/api/buffers.py
asankah: Done.

test/test_vim.py
asankah: Done. D'oh!

neovim/api/nvim.py
```diff
@@ -331,6 +331,53 @@ def handler():
...
+            :py n = len(vim.buffers)    # Number of elements
+            :py for b in vim.buffers:    # Iterating over buffer list
+
+    The key for indexing into vim.buffers[] is the buffer number as described
+    in *python-buffer*.
+    """
+
+    def __init__(self, nvim):
+        """Initialize a Buffers object with Nvim object `nvim`."""
+        self._fetch_buffers = functools.partial(nvim.request,
```
bfredl: you can write this shorter as `nvim.api.get_buffers`

asankah: Done.

asankah:
Ready for another pass (Not sure how GitHub code reviews work).

bfredl:
Thanks, I will take another look tonight, it is probably ready for merge.

neovim/api/nvim.py
```diff
@@ -331,6 +331,52 @@ def handler():
...
 
+class Buffers(object):
+
+    """Remote NVim buffers.
+
+    Currently the interface for interacting with remote NVim buffers is the
+    `vim_get_buffers` msgpack-rpc function. Most methods fetch the list of
+    buffers from NVim.
+
+    Conforms to python-buffers, which reads:
```
bfredl: I don't think we need to cite this verbatim, a reference to `*python-buffers*` is enough.

asankah: Done.

neovim/api/nvim.py
```diff
@@ -331,6 +331,52 @@ def handler():
...
+    def __len__(self):
+        """Return the count of buffers."""
+        return len(self._fetch_buffers())
+
+    def __getitem__(self, number):
+        """Return the Buffer object matching buffer number `number`."""
+        for b in self._fetch_buffers():
+            if b.number == number:
+                return b
+        raise KeyError("No buffers match buffer number {}".format(number))
```
bfredl: `KeyError(number)` is enough

asankah: Done.

bfredl:
Thanks, merged!

asankah:
Woo!

ianks:
Has this been released? I have a similar error with `:YcmComplete RefactorRename` after upgrading to the newest client via pip.

asankah:
@ianks : Which version? The change should be in 0.1.9.

justinmk:
In the latest HEAD version of neovim `:CheckHealth!` is a nice way to see the state of your python env.

dylanscott:
@ianks I was seeing a similar issue and `:CheckHealth` revealed that while my python3 neovim-client was at 0.1.9, the python2 one was at 0.1.7. Not sure if having both installed is bad but upgrading the python2 client fixed `:YcmComplete RefactorRename`

PR #221, opened 2016-09-14
Add exc to first line of ErrorResponse message

blueyed:
`v:exception` only contains the first line, and therefore the real error will not be visible in Neovim, e.g. `PermissionDenied`.
Adding it to the first line of the message helps.  This is https://github.com/neovim/python-client/commit/194317d9490962b0e69a590623d49af938f968ad.

And then I've done it in other places, too - but without testing it.

### DIFF
--- a/pynvim/api/nvim.py
+++ b/pynvim/api/nvim.py
@@ -208,7 +208,7 @@ class Nvim(object):
             args = walk(self._from_nvim, args)
             try:
                 result = request_cb(name, args)
-            except Exception as exc:
+            except Exception:
                 self._err_cb("error caught in request handler '{} {}': {}"
                              .format(name, args, format_exc_msg()))
                 raise
@@ -219,7 +219,7 @@ class Nvim(object):
             args = walk(self._from_nvim, args)
             try:
                 notification_cb(name, args)
-            except Exception as exc:
+            except Exception:
                 self._err_cb("error caught in notification handler '{} {}': {}"
                              .format(name, args, format_exc_msg()))
                 raise
@@ -443,7 +443,7 @@ class Nvim(object):
         def handler():
             try:
                 fn(*args, **kwargs)
-            except Exception as err:
+            except Exception:
                 msg = ("error caught while executing async callback:\n"
                        "{}the call was requested at\n{}"
                        .format(format_exc_msg(), call_point))
--- a/pynvim/api/nvim.py
+++ b/pynvim/api/nvim.py
@@ -208,9 +208,9 @@ class Nvim(object):
             args = walk(self._from_nvim, args)
             try:
                 result = request_cb(name, args)
-            except Exception:
-                msg = ("error caught in request handler '{} {}'\n{}\n\n"
-                       .format(name, args, format_exc_skip(1)))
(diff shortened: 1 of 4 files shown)


bfredl:
Hmm, this only prints the exception message, not the exception typename, maybe we want "{}: {}" with args `type(exc).__name__, exc`.

But it seems also `v:exception` problem in general, consider (in vim)
```
try
python3 do_something_causing_error()
catch
echo v:exception
endtry
```
here `v:exception` is also absolutely useless, and I reckon it would be for much user code using multiline `errmsg`. It seems like a design error in vim that it is not possible to get the complete error output between try and catch. (well I guess it might be, but I mean with an obvious `v:` variable)

justinmk:
@blueyed ping

blueyed:
@justinmk 
pong.

Rebased it, but am not sure what the actionable here is.

@bfredl 
> Hmm, this only prints the exception message, not the exception typename, maybe we want "{}: {}" with args type(exc).__name__, exc.

Should I change it to include the typename?

cherti:
@blueyed 

> Should I change it to include the typename?

I think @bfredl's intention was that you not only would see the error message, but also the error type, therefore I would also understand this suggestion the way you said.

justinmk:
@blueyed could you add a test?

blueyed:
I agree that it would need tests, especially since no existing ones are failing.
Might take a while though.

blueyed:
Even with these improvements I think some additional `v:exception_info` might be useful, which could be a dict, which could then hold something like `python_traceback`, containing the lines of the traceback.

This should also be discussed / get aligned with Vim probably.

ZyX-I:
You may want to look at neovim/neovim#7969, it is supposed to allow multiline exceptions which appear nicely, so that you could just pump the whole traceback as one big error message. `v:exception` for lua errors is supposed to work fine (i.e. contain the whole error) with and without neovim/neovim#7969, PR only changes appearances when uncaught. Note that in this case it is possible to keep some backwards compatibility: the first line will still be the first line, so anybody matching it without including `$` anchor will match new multiline exceptions as well.

blueyed:
@ZyX-I 
Thanks for the pointer, will check it out.

@justinmk 
Haven't tried, likely not.
I'm not eager to get this merged quickly, but rather want to test-drive it a bit more myself and/or would appreciate comments on the code / approval of the PR itself first.

justinmk:
@blueyed A test might give me something to comment on, otherwise LGTM.

blueyed:
I think it would be best to support multiline error messages in Neovim/python also (https://github.com/neovim/neovim/pull/9520 (https://github.com/neovim/neovim/pull/7969)).
Changes here still make sense in general, but I just came back to it thinking about moving the (Python) error location also to the first line to work around this missing feature.
Any pointers what's needed on Neovim's side?

wookayin:
Since neovim v0.4.0 `v:exception` is a multi-line string and therefore contains all the stacktrace information (the upstream issue was fixed), so I think we'd no longer this.

(Sorry for bumping up a 3-year-old issue; I was touching on a codebase that has related to exception handling, and it now seems it can be closed)

PR #223, opened 2016-09-23
logging: use predictable log file names

### LINKED ISSUE #222: Predictable logfile name
In https://github.com/neovim/python-client/blob/b8bb7c21f955fffb8c48ced31158772911604770/neovim/__init__.py#L101-L102 the value of `NVIM_PYTHON_LOG_FILE` is used to build the log file, but the python-client's PID gets appended.
That is unpredictable, e.g. when you just want to use `tail -F`, and you are likely to restart Neovim even.

You can use tricks like `logfile_*(om[1])` in zsh to use the most recent file, but then it won't be followed by `tail` when restarting Neovim.

I think there should be a way to just specify the logfile directly (without changing the source code).

bfredl:
Fixes #222 . Logs now have names as `{$NVIM_PYTHON_LOG_FILE}_py{X}_{NAME}` where X is 2 or 3 and NAME is either "rplugin" or "script" (or "test" when running nosetests with logging). ping @blueyed

### DIFF
--- a/neovim/__init__.py
+++ b/neovim/__init__.py
@@ -53,7 +53,14 @@ def start_host(session=None):
         if os.path.isdir(path) and dup in plugins:
             plugins.remove(dup)
 
-    setup_logging()
+    # Special case: the legacy scripthost receives a single relative filename
+    # while the rplugin host will receive absolute paths.
+    if plugins == ["script_host.py"]:
+        name = "script"
+    else:
+        name = "rplugin"
+
+    setup_logging(name)
 
     if not session:
         session = stdio_session()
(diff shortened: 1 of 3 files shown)


neovim/__init__.py
```diff
@@ -94,12 +101,13 @@ def attach(session_type, address=None, port=None,
...
-def setup_logging():
+def setup_logging(name):
     """Setup logging according to environment variables."""
     logger = logging.getLogger(__name__)
     if 'NVIM_PYTHON_LOG_FILE' in os.environ:
-        logfile = (os.environ['NVIM_PYTHON_LOG_FILE'].strip() +
-                   '_' + str(os.getpid()))
+        prefix = os.environ['NVIM_PYTHON_LOG_FILE'].strip()
+        major_version = sys.version_info[0]
+        logfile = '{}_py{}_{}'.format(prefix, major_version, name)
```
justinmk: Isn't positional formatting required in some python versions? E.g. {0} {1}

bfredl: True, I just assumed no one would ever debug python plugins on 2.6. But then let's make it properly, release 0.1.10 with the HEADSUP that it is the last version supporting 2.6, and then formally remove 2.6 support.

blueyed:
Isn't there some doc about this somewhere (which would need to be changed)?
Otherwise there should be some probably?!

bfredl:
Noted, updated `Readme.md`

PR #224, opened 2016-09-29
Allow defining a group for the autocmd decorator

jamessan:


neovim/plugin/decorators.py
```diff
@@ -83,7 +83,7 @@ def dec(f):
     return dec
 
 
-def autocmd(name, pattern='*', sync=False, eval=None):
+def autocmd(name, group=None, pattern='*', sync=False, eval=None):
```
fwalch: Maybe add `group` as the last argument to avoid a breaking change (if someone called the function without keyword arguments)?

bfredl:
Does this actually work without a change to nvim as well, in `runtime/autoload/remote/define.vim`

```vim
function! remote#define#AutocmdOnHost(host, method, sync, name, opts)
  let group = s:GetNextAutocmdGroup()
  let forward = '"doau '.group.' '.a:name.' ".'.'expand("<amatch>")'
  let a:opts.group = group
  let bootstrap_def = s:GetAutocmdPrefix(a:name, a:opts)
        \ .' call remote#define#AutocmdBootstrap("'.a:host.'"'
        \ .                                ', "'.a:method.'"'
        \ .                                ', '.string(a:sync)
        \ .                                ', "'.a:name.'"'
        \ .                                ', '.string(a:opts).''
        \ .                                ', "'.escape(forward, '"').'"'
        \ .                                ')'
  exe bootstrap_def
endfunction
```

it looks like `opts.group` always overridden by `NextAutocmdGroup`

jamessan:
> it looks like `opts.group` always overridden by `NextAutocmdGroup`

Thanks for the reminder.  I had noticed that while looking into this, but then forgot about it as I was poking around at other things.

That behavior has existed since it was merged to neovim/master.  Looking through the various PRs around developing providers/rplugins, I don't see any discussion around __why__ that decision was made.

bfredl:
Maybe it was not simply not intended for the group to be setable, the `v:opts` dictionary was just reused to convey this value internally in nvim (instead of needing to add a separate "group" parameter to all constructed `:execute` calls). IIRC autocommand groups are really mostly to handle redefinitons if vimrc or a plugin.vim file gets re-sourced. But I suppose they also serve a documentary purpose if the user lists all active autocommands? Maybe the group name should be autogenerated from the rplugin name rather than useless `RPC_DEFINE_AUTOCMD_GROUP_{arbitrary number}`.

jamessan:
> Maybe the group name should be autogenerated from the rplugin name rather than useless RPC_DEFINE_AUTOCMD_GROUP_{arbitrary number}.

That sounds better.

PR #232, opened 2016-10-24
[RFC] update to api level 1

bfredl:
Citing the nvim help:

> The Python client is the reference implementation for API clients. It is
always up-to-date with the Nvim API, so its source code and test suite are
authoritative references.

As the tests show, this is a slightly _breaking_ change for `obj.api.method(...)`. We could mitigate that by letting the `RemoteApi` contain a whitelist of deprecated and renamed methods, but I doubt that it would affect any plugin as all the functionality of the deprecated and renamed methods are more conveniently exposed through the standard (hard-coded) python interface.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ if platform.python_implementation() != 'PyPy':
     install_requires.append('greenlet')
 
 setup(name='neovim',
-      version='0.1.10',
+      version='0.1.11dev',
       description='Python client to neovim',
       url='http://github.com/neovim/python-client',
       download_url='https://github.com/neovim/python-client/archive/0.1.10.tar.gz',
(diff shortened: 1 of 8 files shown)


bfredl:
Also added proposal for exposing versions and version check for recent enough nvim.

setup.py
```diff
@@ -19,7 +19,7 @@
     install_requires.append('greenlet')
 
 setup(name='neovim',
-      version='0.1.10',
+      version='0.1.11dev',
```
bfredl: No, `setup.py` can't import the package currently being installed. If anything, information should flow in the other direction, but I don't know how...

blueyed: setuptools_scm?
See https://github.com/pytest-dev/pytest-mock/pull/67/files for an example integration.

bfredl: Thanks for the pointer, I will look into it tomorrow maybe

bfredl: looks useful, but could be added later on.

PR #233, opened 2016-10-28
Add number attribute to Window and Tabpage objects

### LINKED ISSUE #87: Window and tabpage objects are missing number attribute

jamessan:
The legacy if_pyth interface exposed a number attribute which was
equivalent to the winnr() and tabpagenr() values for the window and
tabpage, respectively.

Neovim 0.1.6 added the necessary API functions to support exposing this
information.

Closes #87


bfredl:
Nice, merged!

PR #234, opened 2016-10-31
remove python 2.6 support

bfredl:


neovim/__init__.py
```diff
@@ -118,7 +118,7 @@ def setup_logging(name):
     if 'NVIM_PYTHON_LOG_FILE' in os.environ:
         prefix = os.environ['NVIM_PYTHON_LOG_FILE'].strip()
         major_version = sys.version_info[0]
-        logfile = '{0}_py{1}_{2}'.format(prefix, major_version, name)
+        logfile = '{}_py{}_{}'.format(prefix, major_version, name)
         handler = logging.FileHandler(logfile, 'w')
```
justinmk: Are the positional specifiers not considered a good practice / idiomatic?

bfredl: I don't get it. There is no specifiers on the other side, so what are the numbers references to? All 2.7+ code bases I have seen do not use them. (OTOH what could be useful in some cases are _string_ specifiers into a dict)

fwalch: > There is no specifiers on the other side, so what are the numbers references to?

The position of the arguments, i.e. `'{0}_{1}_{0}'.format('foo', 'bar')` will give `foo_bar_foo`.

bfredl: I didn't mean I didn't _understand_ what it it means, just that it makes no _sense_ . If the reader is able to count the arguments on the right side, then the reader should be capable to do that on the left side as well. If a case is so complicated that cues are needed, then those should be on _both_ sides, using named arguments.

bfredl: It will get worse, python 3.6 will have a dedicated `f"format"` string literal kind where you directly can access locals :)

fwalch: I guess it can be useful in certain situations, e.g. something like:
```
x = ['a', 'b', 'c', 'd']
'{2}'.format(*x)

'real: {0.real}, imag: {0.imag}'.format(complex(2,3))
```

But yeah:

> Python has too many ways to do things.

But anyway, :+1: for this change, reads better without the numbers IMO.

justinmk:
We might need to add a check for python 2.7 in CheckHealth, if there isn't already.

PR #236, opened 2016-11-09
api: export Object.handle and make Buffer.number non-blocking

bfredl:
In conversion with @justinmk (on gitter), we decided to "freeze" the current msgpack EXT representation objects, which simply is the handle (buffer number, window id) of the object. This means `nvim_buf_get_number` is very unneccesery, it takes the (doubly-encoded) number and just echoes the same number back, single-encoded.

This makes `buffer.number` deduce the number from the `code_data`, instead of using a blocking rpc call. This means it can be safely used in async handlers and threads, for instance buffer change notifications (neovim/neovim#5269) will send a buffer object in async notifications, but one might want to use the number for bookkeeping etc.

This also exposes Object.handle, I think it is very unlikely that vim will introduce something conflicting here, if anythink they might add `Window.winid` which we can alias the same way.

### DIFF
--- a/neovim/api/common.py
+++ b/neovim/api/common.py
@@ -1,6 +1,8 @@
 """Code shared between the API classes."""
 import functools
 
+from msgpack import unpackb
+
 from ..compat import unicode_errors_default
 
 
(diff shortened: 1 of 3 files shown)


neovim/api/common.py
```diff
@@ -21,6 +23,7 @@ def __init__(self, session, code_data):
         """
         self._session = session
         self.code_data = code_data
+        self.handle = unpackb(code_data[1])
```
phodge: You could add `self.number = self.handle` here and the @property `number()` method above could be removed altogether.

bfredl: It won't work as windows has `number` different from `handle`. It would require a constructor for `Buffer` and I think keeping the property is simpler.

PR #241, opened 2017-01-07
Attach child handler loop.

meiralins:
Fix problems reported in #238 for Python3.6. ( In python3.5 I couldn't reproduce the error)
Tested with python versions: 2.7.13, 3.4.5, 3.5.2 and 3.6.0.

In python 3.4, and 3.5, we have this:
```
Exception ignored in: <bound method _UnixSelectorEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
Traceback (most recent call last):
  File "/home/meiralins/.pyenv/versions/3.4.5/lib/python3.4/asyncio/base_events.py", line 417, in __del__
  File "/home/meiralins/.pyenv/versions/3.4.5/lib/python3.4/asyncio/unix_events.py", line 58, in close
  File "/home/meiralins/.pyenv/versions/3.4.5/lib/python3.4/asyncio/unix_events.py", line 139, in remove_signal_handler
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
Exception ignored in: Exception ignored in: Exception ignored in: %
```

```
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
Traceback (most recent call last):
  File "/home/meiralins/.pyenv/versions/3.5.2/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/home/meiralins/.pyenv/versions/3.5.2/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/home/meiralins/.pyenv/versions/3.5.2/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/home/meiralins/.pyenv/versions/3.5.2/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

I'm not sure how to fix it yet.

PR #242, opened 2017-01-10
Add tox

meiralins:


meiralins:
So, this PR tests python3.6 in travis and now I guess we could bump a new version given #241 is necessary to run on python3.6.

justinmk:
Is it necessary to run the "flake" build on python 2 and 3? Perhaps we only need the "flake" build on python 3. That would eliminate one of the builds.

meiralins:
I don't think it is. Django for example only uses flake on python3. [django/tox.ini](https://github.com/django/django/blob/master/tox.ini)
I can change it to tox also.

justinmk:
I guess just this needs to be deleted from travis.yml:

```
   include:
     - python: 2.7
       env: CI_TARGET=flake
```

justinmk:
LGTM. Ready?

bfredl:
> we could bump a new version given #241 is necessary to run on python3.6.

:+1: I will release a version after this PR.

meiralins:
:shipit:

PR #258, opened 2017-05-11
Implementation of global function 'nvim_err_writeln'

stsewd:
Hi, the global function `nvim_err_writeln` was missing, so I added it. I took the documentation from the [neovim docs](https://github.com/neovim/neovim/blob/master/runtime/doc/api.txt#L295).

Sorry I haven't added any test, I don't know how to test this functionality and I didn't find any similar test.


bfredl:
it does not really provide any functionally, it behaves the same as `err_write` but just adds a `\n` afterwards. One could just use `err_write` and add `\n` oneself.

PR #270, opened 2017-10-10
Raise end-index by 1 when setting parts of Range

cherti:
This commit fixes a bug in Range, leading to the last line of a Range
not being able to be replaced via `__setitem__` as the final
line setting its contents in the buffer the Range is based on assumes the
end-index to be inclusive, which it isn't in Python-slicing-notation.

@justinmk

### DIFF
--- a/neovim/api/buffer.py
+++ b/neovim/api/buffer.py
@@ -160,8 +160,8 @@ class Range(object):
         if start is None:
             start = self.start
         if end is None:
-            end = self.end + 1
-        self._buffer[start:end] = lines
+            end = self.end
+        self._buffer[start:end + 1] = lines
 
     def __iter__(self):
         for i in range(self.start, self.end + 1):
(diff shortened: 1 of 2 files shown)


cherti:
I checked the failed checks of CI, it's flake complaints unrelated to this PR, therefore I'd say this should be ready for merge nevertheless.

bfredl:
could a test be added to `test/test_buffer.py` ?

cherti:
Certainly so. Done.

cherti:
PR is updated, `__setitem__` should now work as intended.

One thing that's still bugging me: I cannot reproduce a similar behavior for `__getitem__`, but as `__getitem__` and `__setitem__` are implementet analogously, I would suspect a similar issue in there.
I cannot reproduce that, though, everything looks fine for `__getitem__`, according to my tests.

cherti:
@bfredl 
ok, cannot reproduce any problems with `__getitem__`, just seems to be with `__setitem__`.

So if the test is sufficient for you, the PR is final and ready for merge.

PR #271, opened 2017-10-11
Ignore D401-type-complaints of flake8

cherti:
Travis-CI-Build currently fails on the codebase of `python-client` in general for all PRs (except this one fixing the issue), because D401 of `flake8` complains about D401 at certain positions.
This adds D401 to the ignore-list in `setup.cfg` to avoid this.


bfredl:
Thanks!

PR #273, opened 2017-10-13
[WIP] [RFC] feature: add notification handler for python_execute

roxma:
This PR is for supporting https://github.com/roxma/nvim-asyncscript

However, there's an issue with this PR, I need some help:

Here are the steps to reproduce the issue:

1. Print a very long string with `call ascript#python3("vim.command('echo &rtp')")`

2. Press `CTRL-C` to interrupt the command

The command line shows error message:

```
During handling of the above exception, another exception occurred:
Press ENTER or type command to continue
```

3. Print another very long string with `call ascript#python3("vim.command('echo &rtp')")`

Neovim says the channel is dead.

```
Error detected while processing function ascript#python3:
line    6:
E475: Invalid argument: Channel doesn't exist
```

### DIFF
--- a/neovim/plugin/script_host.py
+++ b/neovim/plugin/script_host.py
@@ -84,6 +84,7 @@ class ScriptHost(object):
 
     @rpc_export('python_execute_async', sync=False)
     def python_execute_async(self, *args):
+        """Same as python_execute, except it's notification handler."""
         try:
             self.python_execute(*args)
         except Exception:
@@ -103,6 +104,7 @@ class ScriptHost(object):
 
     @rpc_export('python_execute_file_async', sync=False)
     def python_execute_file_async(self, *args):
+        """Same as python_execute_file, except it's notification handler."""
         try:
             self.python_execute(*args)
         except Exception as ex:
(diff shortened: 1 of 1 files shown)


roxma:
Although the exception is raised wile executing the method, I cannot catch it with `try: ... except:` block.

I don't understand why.

It seems to be related to the greenlet and async event loop feature.

roxma:
There's a disavantange with current implementation,

It's hard to pass local vimscript variables to python without pushing them to the global. With global variables, strictly speeking, async job has to take care of synchronization. If so, it still makes this feature hard to use.

bfredl:
There is an existing alternative: `:py3 vim.async_call("function", ars)`. It is slightly less efficient, but you can do sync requests in `args`.

bfredl:
Yet another alternative would be to add `Python3call("expr", args, async)` where args are evaluated vimL side.

roxma:
@bfredl  

Thank you very much! I have switched to the easier implementation. There's no need for merging this PR. I'm closing this PR.

PR #274, opened 2017-10-13
Fix 'SyntaxError: invalid syntax' on Python 3.7

lambdalisue:
1__tmux

### Reason

'async' is keyword in Python 3.7

https://www.python.org/dev/peps/pep-0492/#deprecation-plans


neovim/api/buffer.py
```diff
@@ -98,17 +98,17 @@ def range(self, start, end):
...
-                      col_end=-1, src_id=-1, async=None):
+                      col_end=-1, src_id=-1, _async=None):
         """Add a highlight to the buffer."""
-        if async is None:
-            async = (src_id != 0)
+        if _async is None:
+            _async = (src_id != 0)
         return self.request('nvim_buf_add_highlight', src_id, hl_group,
-                            line, col_start, col_end, async=async)
+                            line, col_start, col_end, _async=_async)
```
justinmk: Does the parameter really need to be changed? That's silly, python parser should know the difference. 

This will be a breaking change for our users

justinmk:
I guess we need to detect the python version and show a warning.

neovim/api/buffer.py
lambdalisue: I machinery replaced all `async` into `_async` so I'm not sure about this particular line.
However the screenshot I pasted showed `invalid syntax` on function attributes and this PR fixed that issue.

lambdalisue:
While Python 3.7 is in alpha stage yet, this PR may not required in future.
I made this PR just for note and users who need Python 3.7 support right now.

bfredl:
So lets wait until python 3.7 is at least in beta. Hopefully they will gather their senses and make it a context-dependent keyword (the pattern used by python in the past). If not, we must keep the existing `async` keyword for <=3.6 and and use something else for >=3.7, perhaps using a script as part of the build system.

meiralins:
I guess we could add a decorator that receives `_async` and `**kwargs`, and handle the argument change for python<3.7.
But eventually, plugins will be updated because they will also break when users update to python3.7.

Looks like a sane approach to me.

meiralins:
Just to make things clear. The approach I propose is that every function with `async` as an argument would accept both `async` and `_async` as arguments via a simple decorator that gets the `async` argument via `**kwargs` and calls the function with the proper `_async` argument. We would simply use `_async` in the lib internals and it would be ok with python3.7. Every lib that explicitly uses the `async` keyword would work in Python<3.7.

When a user upgrades to python3.7, every lib that uses explicitly the `async` keyword would fail and there's nothing we can do about it. Eventually, the lib maintainers would need to fix the problem. At some point we could even delete the decorator.

What do you guys think about it @justinmk @bfredl @lambdalisue ?

@lambdalisue if those guys are okay with that approach feel free to implement it.

lioumens:
Thanks for the PR @lambdalisue. FYI python 3.7 is now in beta.

bfredl:
Is the breaking change still there? I looked into  https://docs.python.org/3.7/whatsnew/3.7.html and grepped for both "async" and "keyword", and found nothing.

Shougo:
@bfredl https://docs.python.org/3.6/whatsnew/3.6.html#new-keywords

It is added in `Python 3.6` what's new documentation.

> New Keywords
> async and await are not recommended to be used as variable, class, function or module names. Introduced by PEP 492 in Python 3.5, they will become proper keywords in Python 3.7. Starting in Python 3.6, the use of async or await as names will generate a DeprecationWarning.

Note: It should be added in `Python 3.7` what's new though.

justinmk:
Looks like other popular projects deal with this in the same way. Though `async_` (underscore at end instead of start), is more common, so let's also use `async_`.

jdemilledt:
Python 3.7.0b2 still has this issue.

lambdalisue:
Shall I replace _async to async_?

justinmk:
@lambdalisue yes, thanks.

lambdalisue:
alacritty 2018-04-13 06-29-28

Done. It seems [tests are failing on `master` as well](https://travis-ci.org/neovim/python-client/builds/365812611) so I just ended-up.

Additionally, you said

> I guess we need to detect the python version and show a warning.

So I added 6d1b28b424b73c945036f47b190a8e873b859e84 for backward compatibility. Please revert whenever you satisfied the new API.

neovim/msgpack_rpc/session.py
```diff
@@ -72,12 +73,19 @@ def request(self, method, *args, **kwargs):
...
+        is sent instead. This will never block, and the return value or error
+        is ignored.
         """
-        async = kwargs.pop('async', False)
-        if async:
+        if 'async_' not in kwargs and 'async' in kwargs:
+            warnings.warn(
+                '"async" attribute is deprecated. Use "async_" instead.',
+                DeprecationWarning,
+            )
```
justinmk: This looks like a good solution to me. 

If no one else comments, ping in a few days, then we merge.

jdemilledt:
Looks like AppVeyor is drunk.

justinmk:
Sorry, I broke master when I merged https://github.com/neovim/python-client/pull/290 , so that's why travis is failing. It will be fixed with https://github.com/neovim/neovim/pull/8265

jdemilledt:
@justinmk Ping

neovim/api/buffer.py
```diff
@@ -98,17 +100,34 @@ def range(self, start, end):
         return Range(self, start, end)
 
     def add_highlight(self, hl_group, line, col_start=0,
-                      col_end=-1, src_id=-1, async=None):
+                      col_end=-1, src_id=-1, async_=None,
+                      **kwargs):
         """Add a highlight to the buffer."""
-        if async is None:
-            async = (src_id != 0)
+        if async_ is None and 'async' in kwargs:
```
bfredl: This seems like it can be a helper function:
`async_ = check_async(async_, **kwargs, (src_id != 0))`

neovim/msgpack_rpc/event_loop/uv.py
```diff
@@ -13,7 +13,7 @@ class UvEventLoop(BaseEventLoop):
 
     def _init(self):
         self._loop = pyuv.Loop()
-        self._async = pyuv.Async(self._loop, self._on_async)
```
bfredl: invalid substitution.

bfredl:
Other that comments, LGTM.

neovim/api/buffer.py
lambdalisue: Should be? It's just one line and used only here so I don't think it's a good idea to extract a simple code into a helper function.

neovim/msgpack_rpc/event_loop/uv.py
lambdalisue: 👍

lambdalisue:
It seems the highlighted two lines are for private attributes so I renamed these to `_async` in https://github.com/neovim/python-client/pull/274/commits/5718624671c432e69f32b3d911f35cceb6a51e47

alacritty 2018-04-18 00-34-39

I would like to melt that commit into https://github.com/neovim/python-client/pull/274/commits/b7c0d12575a10eb7e1688629d136134281594753 once you satisfied that changes.

neovim/api/buffer.py
bfredl: How so? it is 8-9 lines unrelated to the original meaning of the function in at least three places (here, `clear_highlight`, and `request`)

lambdalisue: I got. Sorry, I thought you are talking about exactly this line.

lambdalisue: Fixed 👍

lambdalisue:
I'll melt 1626e93 into https://github.com/neovim/python-client/pull/274/commits/c421bf58da8ac4527e834560dfa3b490d1784872 later 👍

neovim/msgpack_rpc/event_loop/uv.py
```diff
@@ -13,7 +13,7 @@ class UvEventLoop(BaseEventLoop):
...
+        self.async_ = pyuv.Async(self._loop, self._on_async)
```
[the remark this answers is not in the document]
lambdalisue: Done by https://github.com/neovim/python-client/pull/274/commits/a9a3a705ced30925d22666b5fefbf5f826917ab0 Is it enough?

lambdalisue:
Done.

jdemilledt:
What release is this planned to be included in?

bfredl:
@jdemilledt 3.7 final is due to 15 june, we will have a release well before then. If I get around to it I will finish the lua PR #325 next week, and then do a release.

cormacrelf:
3.7 is out now, and the latest release of neovim doesn't have these fixes.

PR #278, opened 2017-11-05
[Proposal,WIP] change indexing-style in Range-Object-creation

cherti:
This commit changes the indexing-style when creating a Range-object
from vim-style to Python-style, i.e. creating a Range-object spanning
the first three lines of a buffer changes from Range(buffer, 1, 3) to
Range(buffer, 0, 3), making the notation consistent with
Python-style-indexing and the behavior of the Buffer-object, that
already follows Python-style indexing.

Otherwise, addressing lines in a buffer via buffer[a:b] is inconsistent with buffer.range(a:b) as both do not return the same subset of lines, which is confusing and does not follow the principle of least surprise. Furthermore, range itself is afterwards indexed starting with 0, so the first line of `myrange` is `myrange[0]`, adding to the inconsistency.

Upstreams thoughts on the matter?

If this is considered mergeable, I'll take care of restructuring the rest of the range-object to internally work correctly with the changed constructor-notation in `__getitem__` and `__setitem__` and all other relevant cases.

### DIFF
--- a/test/test_buffer.py
+++ b/test/test_buffer.py
@@ -176,6 +176,6 @@ def test_contains():
 @with_setup(setup=cleanup)
 def test_set_items_for_range():
     vim.current.buffer[:] = ['a', 'b', 'c', 'd', 'e']
-    r = vim.current.buffer.range(1, 3)
+    r = vim.current.buffer.range(0, 3)
     r[1:3] = ['foo']*3
     eq(vim.current.buffer[:], ['a', 'foo', 'foo', 'foo', 'd', 'e'])
--- a/neovim/api/buffer.py
+++ b/neovim/api/buffer.py
@@ -134,7 +134,7 @@ class Buffer(Remote):
 class Range(object):
     def __init__(self, buffer, start, end):
         self._buffer = buffer
-        self.start = start - 1
+        self.start = start
         self.end = end - 1
 
     def __len__(self):


test/test_buffer.py
```diff
@@ -176,6 +176,6 @@ def test_contains():
 @with_setup(setup=cleanup)
 def test_set_items_for_range():
     vim.current.buffer[:] = ['a', 'b', 'c', 'd', 'e']
-    r = vim.current.buffer.range(1, 3)
+    r = vim.current.buffer.range(0, 3)
```
justinmk: Won't this break existing plugins that use python-client?

cherti: yes, this would be an API-break to make range generation and buffer-indexing and such consistent overall with python-indexing starting at 0 (and make the buffer[a, b] and buffer.range(a, b) consistent as well in the process.

It is definitely an API-break, that's why I prepended "Proposal" to it because I'd like to know upstreams opinion on this matter.

I think that API-consistency is worth the API-break, but it's not on me to decide that as I am not the maintainer. :)

justinmk: It's not really inconsistent, Range is a Vim concept, see `:help :range`. Buffers start with line 1.

`buffer[a:b]` is treated differently because it's considered a "list" not a "buffer", just like if you had a list of lines in VimL it would be zero-indexed.

cherti: Ah, ok, then the current API makes perfect sense. Thanks for clarifying!

PR #289, opened 2017-12-30
flake8: only disable F821 where needed

blueyed:


neovim/plugin/decorators.py
```diff
@@ -22,7 +22,7 @@ def plugin(cls):
     # decorated functions have a bound Nvim instance as first argument.
     # For methods in a plugin-decorated class this is not required, because
     # the class initializer will already receive the nvim object.
-    predicate = lambda fn: hasattr(fn, '_nvim_bind')
+    predicate = lambda fn: hasattr(fn, '_nvim_bind')  # noqa: E731
```
bfredl: This just makes the code ugly for no reason. If we think named lambda is OK here, why would it be wrong somewhere else?

setup.cfg
```diff
@@ -1,2 +0,0 @@
-[flake8]
-ignore = D211,E731,F821,D401
```
bfredl: I do not think this is safe. It might cause trouble for someone else that has a different set of flake8 subpackages installed. Unless it can be confirmed that `D401` and `D211` was disabled by default in new flake8 version and they have decided to keep it that way.

neovim/plugin/decorators.py
blueyed: I've thought it's a special case here for some reason, but not really read the comment.

So you would like to disable E731 globally then?

setup.cfg
blueyed: I've tried with an empty `ignore` and there were no issues, but would have to investigate then.

blueyed:
Removed 1628f5a.

PR #290, opened 2017-12-30
tests: fix test_command_output

blueyed:
1. The `with_setup` decorator does not appear to be designed to be used
   without parenthesis.  Remove it, since it is not needed/used anyway.
2. Fix the actual test.

### DIFF
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -26,10 +26,8 @@ def test_command():
     os.unlink(fname)
 
 
-@with_setup
 def test_command_output():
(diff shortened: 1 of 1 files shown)


blueyed:
btw: this makes the tests run on pytest.

What do you think about using pytest as main test runner?

justinmk:
> What do you think about using pytest as main test runner

Definitely in favor, the work was done in https://github.com/neovim/python-client/pull/266 but it's blocked by a build failure that no one has investigated.

Issue #295, opened 2018-01-19
DirChanged support


Shougo:
https://github.com/neovim/neovim/issues/3636

neovim-ruby already supports the feature.
https://github.com/alexgenco/neovim-ruby/pull/22/file

PR #296, opened 2018-01-19
Correctly change the working directory in python

Leandros:
This is implementing #295 and https://github.com/neovim/neovim/issues/3636.

### DIFF
--- a/neovim/plugin/script_host.py
+++ b/neovim/plugin/script_host.py
@@ -45,6 +45,14 @@ class ScriptHost(object):
         self.legacy_vim = LegacyVim.from_nvim(nvim)
         sys.modules['vim'] = self.legacy_vim
 
+        # Handle DirChanged. #296
+        nvim.command(
+            'autocmd DirChanged * call rpcrequest({}, "python_chdir", v:event)'
+            .format(nvim.channel_id), async=True)
+        # XXX: Avoid race condition.
+        # https://github.com/neovim/python-client/pull/296#issuecomment-358970531
+        os.chdir(nvim.eval('getcwd()', async=False))
+
     def setup(self, nvim):
         """Setup import hooks and global streams.
 
@@ -152,6 +160,11 @@ class ScriptHost(object):
     def python_eval(self, expr):
         """Handle the `pyeval` vim function."""
         return eval(expr, self.module.__dict__)
+
+    @rpc_export('python_chdir', sync=True)
+    def python_chdir(self, args):
+        """Handle working directory changes."""
+        os.chdir(args['cwd'])
 
     def _set_current_range(self, start, stop):
         current = self.legacy_vim.current
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -8,15 +8,15 @@ Using pip
 
 You can install the package without being root by adding the ``--user`` flag::
 
-    pip2 install neovim
-    pip3 install neovim
+    pip2 install --user neovim
+    pip3 install --user neovim
 
 .. note::
   
     If you only use one of python2 or python3,
     it is enough to install that version.
 
-If you follow Neovim master,
+If you follow Neovim HEAD,
 make sure to upgrade the ``python-client`` when you upgrade Neovim::
 
     pip2 install --upgrade neovim
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-import os, tempfile
+import os, sys, tempfile
 from nose.tools import with_setup, eq_ as eq, ok_ as ok
 from test_common import vim, cleanup
 
@@ -15,7 +15,7 @@ def source(code):
 def test_command():
     fname = tempfile.mkstemp()[1]
     vim.command('new')
-    vim.command('edit %s' % fname)
+    vim.command('edit {}'.format(fname))
     # skip the "press return" state, which does not handle deferred calls
     vim.input('\r')
     vim.command('normal itesting\npython\napi')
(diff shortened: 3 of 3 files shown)


Leandros:
@justinmk Sure, I've had a test written anyway.

neovim/plugin/script_host.py
```diff
@@ -65,7 +65,9 @@ def setup(self, nvim):
         sys.stderr = RedirectStream(lambda data: nvim.err_write(data))
         # Setup the DirChanged listener
         cid = nvim.channel_id
-        nvim.command('au DirChanged * call rpcrequest(%d, "python_chdir", v:event)' % cid)
+        cmd = 'au DirChanged * call rpcrequest(%d, "python_chdir", v:event)' \
```
mhinz: Please use `autocmd` instead of `au`, which makes it easier go grep for when searching for autocmd-related things.

Leandros: Sure.

Leandros:
Alright, all tests ran through. Fixed the last linter warning.

neovim/plugin/script_host.py
```diff
@@ -63,6 +63,12 @@ def setup(self, nvim):
         self.saved_stderr = sys.stderr
         sys.stdout = RedirectStream(lambda data: nvim.out_write(data))
         sys.stderr = RedirectStream(lambda data: nvim.err_write(data))
+        # Setup the DirChanged listener
```
bfredl: Hmm maybe we can use the host autocmd infrastructure here. (We need to add a flag to not start the host for the autocmd, but that would be useable in general)

bfredl: Actually that would require a bit bigger refactor than I thought, so let's keep this as is for now (and maybe revisit at `UpdateRemotePlugins` refactor)

mhinz:
@bfredl 

What is happening here? https://travis-ci.org/neovim/python-client/jobs/330770820#L512

I get this locally as well. Sometimes.

```
Error detected while processing function provider#python3#Call:
line   18:
no request handler registered for "python_execute"
```

It starts happening at a8a98b8.

bfredl:
you mean you don't see it on master?

neovim/plugin/script_host.py
```diff
@@ -63,6 +63,12 @@ def setup(self, nvim):
...
+        cid = nvim.channel_id
```
bfredl: unnecessary temporary

neovim/plugin/script_host.py
```diff
@@ -63,6 +63,12 @@ def setup(self, nvim):
...
+        cmd = 'autocmd DirChanged *' \
+            + ' call rpcrequest(%d, "python_chdir", v:event)' \
```
bfredl: also we prefer to use new-style `.format()`

mhinz:
Yup. I just do:

```
$ git checkout a8a98b8
$ nvim
:python3 import os
```

Commenting out the following line makes it work again:

```
nvim.command('au DirChanged * call rpcrequest(%d, "python_chdir", v:event)' % cid)
```

mhinz:
Any `nvim.command()` in `setup()` will lead to that error.

bfredl:
Hmm, maybe it shouldn't be done in  `__init__`. I generally recommend against doing any RPC there, but I don't  see the specific problem right here. Maybe the issue is that it is un-synchorized in general, using `nvim.command(...,async=True)` might fix it even if its not pretty either.

mhinz:
Yup,`nvim.command(..., async=True)` fixes it for me.

neovim/plugin/script_host.py
Leandros: I, too, prefer `.format()`. I simply followed what I've seen being used before. Will change it.

bfredl:
It is a bit racy, if a vim script does
```
py3 import blah # first call, so starts the host
cd mydir
```
the `cd` might come before the async registration. One can of course also invoke it in the async command to be sure in this case, but generally we want some mechanism for "run this vimL code right when the host starts".

Leandros:
@mhinz Hmm. That is interesting. I'm using it already, and couldn't see any errors. Which nvim version are you using?

mhinz:
@Leandros 

Hm, you're right. I can only reproduce it with master and this repo is using nightlies during CI tests.

I can't reproduce it with the 0.2.2 release.

I'll bisect nvim later.

mhinz:
@bfredl 

This issue starts happening after https://github.com/neovim/neovim/commit/1ebc96fe1.

bfredl:
Yes, it changes processing of RPC job messages,  but not more than to how they were before RPC jobs gained stderr support neovim/neovim#4723 (and to how RPC sockets always have worked), so they are less "forgiving" to potentially racy patterns.

Synchronous initialization should be initiated by one party only. In the case of plugin hosts, this is already nvim, by using one of `poll`, `specs` or in the case of `script_host` by calling one of the script methods directly.

neovim/plugin/script_host.py
```diff
@@ -44,6 +44,14 @@ def __init__(self, nvim):
         exec('import sys', self.module.__dict__)
         self.legacy_vim = LegacyVim.from_nvim(nvim)
         sys.modules['vim'] = self.legacy_vim
+        # Since the 'DirChanged' autocmd is established asynchronously, import
+        # os manually already, to prevent a potential race condition
+        exec('import os', self.module.__dict__)
```
bfredl: Why is this needed? The autocmd handler method is executed in the context of `script_host.py` which already has the import.

Leandros: I tried to work-around the race you described, by importing os before, so that Vim can see the current working directory without creating a race.

Leandros:
@bfredl What's your suggestion of implementing this? It's definitely required, we need to watch for the DirChange somehow. I've just looked at the ruby plugin, and found they're doing it quite similar.

neovim/plugin/script_host.py
bfredl: But the race has nothing to do with `import os`. Quick fix would rather be to do 
`nvim.command('call rpcrequest({}, "python_chdir", {"cwd": getcwd()}', async=True)` after registering the autocmd

bfredl:
Lets use the current solution for now (with the fix I just posted), until we improve host autocmd registration.

neovim/plugin/script_host.py
Leandros: The quickfix doesn't seem to work, at least for me, it results in a python error from my plugins.
screen shot 2018-01-24 at 10 11 03

justinmk:
@Leandros can you include the fix @bfredl mentioned? Would like to move this forward

justinmk:
> `nvim.command('call rpcrequest({}, "python_chdir", {"cwd": getcwd()}', async=True)`

seems to be a chicken-egg problem, instead I went with this:

    os.chdir(nvim.eval('getcwd()', async=False))

Pushed an update to @Leandros branch, will merge after CI.

blueyed:
This appears to cause problems with initial use of `:py3` for me: https://github.com/neovim/python-client/issues/337.

PR #297, opened 2018-01-24
Fix VimPathFinder.find_spec (Python 3.4+)

blueyed:
The fix is similar to what VimModuleLoader.find_module does already.

### DIFF
--- a/neovim/plugin/script_host.py
+++ b/neovim/plugin/script_host.py
@@ -227,8 +227,10 @@ def path_hook(nvim):
 
         def load_module(self, fullname, path=None):
             # Check sys.modules, required for reload (see PEP302).
-            if fullname in sys.modules:
+            try:
                 return sys.modules[fullname]
+            except KeyError:
+                pass
             return imp.load_module(fullname, *self.module)
 
     class VimPathFinder(object):
--- a/neovim/plugin/script_host.py
+++ b/neovim/plugin/script_host.py
@@ -242,9 +242,9 @@ def path_hook(nvim):
                 return None
 
         @staticmethod
-        def find_spec(fullname, path=None, target=None):
+        def find_spec(fullname, target=None):
             """Method for Python 3.4+."""
-            return PathFinder.find_spec(fullname, path or _get_paths(), target)
+            return PathFinder.find_spec(fullname, _get_paths(), target)
 
     def hook(path):
         if path == nvim.VIM_SPECIAL_PATH:


neovim/plugin/script_host.py
```diff
@@ -227,6 +229,13 @@ def find_module(fullname, path=None):
         @staticmethod
         def find_spec(fullname, path=None, target=None):
             """Method for Python 3.4+."""
+            # Check sys.modules, required for importlib.reload.
+            # path appears to be the module then, and calling find_spec with it
+            # fails.
+            try:
+                return sys.modules[fullname].__spec__
```
blueyed: `__spec__` might be None here.. seen when using jedi-vim, which apparently also comes here.

Need to investigate..

blueyed: @davidhalter
Any idea maybe?
Maybe AttributeError could be handled like KeyError, but it would be good to know what is going on.

I do not have the details anymore, but am using this branch again now (to fix `reload(jedi_vim)` in Neovim).

blueyed:
While https://github.com/neovim/python-client/pull/297#discussion_r163708678 is not really clear, I think this PR is better than the current situation, where `reload()` does not work at all:

`:py3 from importlib import reload; reload(jedi_vim)` results in:

```
Error detected while processing function provider#python3#Call:
line   18:
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/importlib/__init__.py", line 165, in reload
    spec = module.__spec__ = _bootstrap._find_spec(name, pkgpath, target)
  File "<frozen importlib._bootstrap>", line 894, in _find_spec          
  File "<frozen importlib._bootstrap_external>", line 1157, in find_spec
  File "<frozen importlib._bootstrap_external>", line 1129, in _get_spec
  File "…/neovim-python-client/neovim/plugin/script_host.py", line 247, in find_spec
    return PathFinder.find_spec(fullname, path or _get_paths(), target)
  File "<frozen importlib._bootstrap_external>", line 1157, in find_spec
  File "<frozen importlib._bootstrap_external>", line 1123, in _get_spec
TypeError: 'module' object is not iterable                              
```

`find_spec` gets called with the following with `reload`:

> find_spec: fullname='jedi_vim', path=<module 'jedi_vim' from '/home/user/.vim/plugged/jedi-vim/pythonx/jedi_vim.py'>, target=None

blueyed:
Found the issue: it was not using the correct interface (`path` is only used with meta path finders!).

justinmk:
LGTM. The flake8 failure happens on master, seems like a flake8 bug. 

flake8 version:

    checkqa installed: flake8==3.5.0

Did a web search but didn't find much.

blueyed:
@justinmk 

> The flake8 failure happens on master, seems like a flake8 bug.

Could you reproduce it locally?
`tox -e checkqa` works for me.

PR #298, opened 2018-02-01
Check count for 'not None' instead of truthy in command decorator

malramsay64:
The command decorator was only checking the truthyness of `count`, which
meant that a value of 0 was considered falsy and so the value not passed
to the decorated function. A value of 0 for `count` should be valid
since this is the default in both vim and neovim (see `help v:count`).


justinmk:
Good catch. Can you add a test?

malramsay64:
There are no other tests I could find for the decorators so I created a new file with the test case, I hope that is alright.

PR #304, opened 2018-03-01
MsgpackStream: fix typos

mhinz:


mhinz:
Looks like a refactoring gone wrong. :-)

I also wasn't sure if you really wanted `_loop` instead of `loop` in the first place.

mhinz:
Guess this needs another release now since PyPI doesn't allow to overwrite `0.2.2`? Not sure.

PR #305, opened 2018-03-04
[WIP] rename package and PyPI package to pynvim

bfredl:
And add dummy transition package.


mhinz:
What's the rationale behind this change? I guess it's because we generally want to call it `nvim` instead of `neovim`? 

Isn't the `py` superfluous?

bfredl:
> Isn't the py superfluous?

No. The point is `py-` allows us to use the _same_ name regardless of "it being a python package" is an implied context or not. There are lot's of precedents for python packages with "py" such as numpy/scipy etc.

Right now, the same software is called `python-client` in neovim contexts, and "neovim" in python contexts. I've heard several reports on gitter "pip install neovim" being confusing for newcomers, as it looks like they install neovim twice. 

See also `:help dev-api-client`

dummy/neovim/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Python client for Nvim.
+
+Client library for talking with Nvim processes via it's msgpack-rpc API.
```
justinmk: "its"

dummy/setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import setup
+
+setup(name='neovim',
+      version='0.2.4dev',
+      description='Transition packgare for pynvim',
```
justinmk: "package"

dummy/neovim/__init__.py
bfredl: Should probably change this line anyway.

PR #312, opened 2018-03-15
Fix deprecated API names in pydoc.

FGtatsuro:
This is a trivial fix. 
I use new API names instead of deprecated ones in pydoc of RemoteSequence.
Ref: https://github.com/neovim/neovim/blob/master/src/nvim/api/dispatch_deprecated.lua

best regards.


bfredl:
Thanks!

PR #313, opened 2018-03-17
Remove dead code.

FGtatsuro:
I found several dead code in neovim/api/common.py. This PR removes them.

Best regards.

### DIFF
--- a/neovim/api/common.py
+++ b/neovim/api/common.py
@@ -69,8 +69,8 @@ class RemoteMap(object):
     It is used to provide a dict-like API to vim variables and options.
     """
 
-    def __init__(self, obj, get_method, set_method=None, self_obj=None):
-        """Initialize a RemoteMap with session, getter/setter and self_obj."""
+    def __init__(self, obj, get_method, set_method=None):
+        """Initialize a RemoteMap with session, getter/setter."""
         self._get = functools.partial(obj.request, get_method)
         self._set = None
         if set_method:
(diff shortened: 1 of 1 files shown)


neovim/api/common.py
```diff
@@ -154,10 +150,6 @@ def __contains__(self, item):
         return item in self._fetch()
 
 
-def _identity(obj, session, method, kind):
-    return obj
```
justinmk: This is for debugging I think

FGtatsuro: Oh, sorry. Thus, I remove this commit.

FGtatsuro: Force updated.

neovim/api/common.py
```diff
@@ -69,8 +69,8 @@ class RemoteMap(object):
     It is used to provide a dict-like API to vim variables and options.
     """
 
-    def __init__(self, obj, get_method, set_method=None, self_obj=None):
-        """Initialize a RemoteMap with session, getter/setter and self_obj."""
```
justinmk: I'm not sure if this is used or not, but it seems like there's no test coverage. If it is used a test would be nice.

FGtatsuro: Thank you for your comment. As you say, I should add a test covering this change. 
I'll add several tests in test/test_common.py.

justinmk: Just one test is enough. @bfredl would know better whether `self_obj` is still needed.

bfredl: `self_obj` looks completely dead to me.

FGtatsuro: @bfredl 

Thank you for your check!

I've checked existing test code(under `test` directory), and I have a question.

`RemoteSequence`/ `RemoteMap` isn't public, and their `__init__` methods(fixed by this PR) are called internally. On the other hand, it seems that current test code covers public API. 
I wonder whether it's better to also cover private APIs(In this case, `__init__`  of `RemoteSequence`/ `RemoteMap`). 

Best regards.

bfredl: I don't think tests of internal classes are needed, if both classes are already covered by public API, there is no need to add a test.

FGtatsuro: OK, Thank you or your comment!

PR #317, opened 2018-03-23
document workflow for local plugin development

### LINKED ISSUE #316: What's the easiest way to develop a new plugin?
Hi, thanks again for helping me out here https://github.com/neovim/python-client/issues/279

I finally found some free time to work on my plugin. Unfortunately I'm struggling with a good development workflow.

I added the file as you suggested: `plugin/my-thing.vim` and I'm starting neovim as `nvim -c ":set runtimepath+=$PWD"` which works great because it loads my plugin from `rplugin/python3/my_plugin/`, but unfortunately it doesn't load the code from the `plugin/` folder. Is this a bug? Is there a better way?

TomasTomecek:
Fixes #316

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -98,7 +98,7 @@ html_theme = 'sphinx_rtd_theme'
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []
 
 
 # -- Options for HTMLHelp output ------------------------------------------
--- a/docs/usage/remote-plugins.rst
+++ b/docs/usage/remote-plugins.rst
@@ -1,4 +1,5 @@
 .. _remote-plugins:
+
 Remote (new-style) plugins
 ==========================
 


bfredl:
This should probably be in `docs/usage/remote-plugins.rst`

Though I wonder if recommending a dedicated vimrc is a bit overblown. Most plugin managers can add a explicit path, or one could just
```
set rtp+=~/dev/nvim-myplugin/
```
in the normal `init.vim`

justinmk:
Though I do think the readme should have a self-contained quick start sample.

TomasTomecek:
> This should probably be in docs/usage/remote-plugins.rst

Moved.

> Though I wonder if recommending a dedicated vimrc is a bit overblown. Most plugin managers can add a > explicit path, or one could just
> 
> set rtp+=~/dev/nvim-myplugin/
> 
> in the normal init.vim

I think that you don't understand my use case: I want to invoke my plugin while being in my upstream git repository using `nvim ...` without the need of crippling my global vimrc or committing real paths anywhere in the plugin. [The proposed solution](https://github.com/neovim/python-client/issues/316#issuecomment-375512220) by @Shougo does it very nice. If you have a better idea, I would like to hear it.

Also please advise me on what I should do in order to get this PR merged.

docs/usage/remote-plugins.rst
```diff
@@ -45,3 +46,48 @@ but it can make asynchronous requests, i.e. passing ``async=True``.
 
 You need to run ``:UpdateRemotePlugins`` in Neovim for changes in the specifications to have effect.
 For details see ``:help remote-plugin`` in Neovim.
+
+If you are working on your plugin locally, it may be desirable to run the
+plugin from a git repository directly without the need of installing or
+symlinking it.
+The easiest way to do this is to create a dedicated vimrc file just for sake of
+development of your plugin. If you are in a root of the git repo with the
+plugin, consider creating this vimrc file:
```
justinmk: This is rather verbose, it could be 1 sentence.

> For local plugin development, it's a good idea to use an isolated vimrc:

docs/usage/remote-plugins.rst
```diff
@@ -45,3 +46,48 @@ but it can make asynchronous requests, i.e. passing ``async=True``.
...
+
+.. code-block:: console
+
+    $ cat vimrc
+    let &runtimepath.=','.escape(expand('<sfile>:p:h'), '\,')
+
+It will append your current work directory to the runtime path so neovim is
+able to find and load your plugin. You can now invoke neovim:
```
justinmk: "neovim" is spell "Nvim" in technical resources (i.e., almost everywhere except the website).

bfredl:
> I want to invoke my plugin while being in my upstream git repository using nvim ... without the need of crippling my global vimrc or committing real paths anywhere in the plugin. 

I fail to see what is "crippling" with adding the main repository directly to the runtime paths. This is how I manage all developed plugins, and instead use branches to switch between "stable" and "development" if needed, just as with all other vcs managed projects. I even "cripple" my rtp with a pure test rplugin which does crazy stuff in methods, because if a method is not called, a remote plugin by design does nothing. 

To be clear I'm not against adding any text for this usecase, but is shouldn't be worded as the "desirable" or "easiest" for local development.

docs/usage/remote-plugins.rst
TomasTomecek: That's true, I wanted to be precise. Will shorten.

docs/usage/remote-plugins.rst
TomasTomecek: Good catch. I did `s/neovim/Neovim/g` to match rest of the doc.

TomasTomecek:
> I fail to see what is "crippling" with adding the main repository directly to
> the runtime paths. This is how I manage all developed plugins, and instead use
> branches to switch between "stable" and "development" if needed, just as with
> all other vcs managed projects. I even "cripple" my rtp with a pure test
> rplugin which does crazy stuff in methods, because if a method is not called, a
> remote plugin by design does nothing.

Sounds like a good idea to describe in detail in the document (or maybe
write a blog post).

> To be clear I'm not against adding any text for this usecase, but is shouldn't
> be worded as the "desirable" or "easiest" for local development.

I agree this an easier approach. Personally I don't work like this - I instead
prefer to sandbox projects I'm working on.

TomasTomecek:
The CI failure is unrelated:
```
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

ping, what do I have to do in order to get this merged?

docs/conf.py
```diff
@@ -98,7 +98,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []
```
stsewd: Just curious, why did you delete the `_static` dir?

docs/usage/remote-plugins.rst
```diff
@@ -45,3 +46,46 @@ but it can make asynchronous requests, i.e. passing ``async=True``.
...
+
+Then just run command ``:UpdateRemotePlugins`` and your plugin should be up and
+running.
+
+In case you run into some issues, you can list loaded plugins:
+
+.. code-block:: console
+
+    :scriptnames
+      1: ~/path/tp/your/plugin-git-repo/vimrc
```
stsewd: This should be `~/path/to/...` I guess, right?

docs/conf.py
TomasTomecek: Because you are not using it and is causing warnings during build of documentation:

https://github.com/neovim/python-client/pull/317/commits/9e8a9d07a02f7ac3a82fcccc6919f3414611da0d

```
WARNING: html_static_path entry u'docs/_static' does not exist
```

docs/usage/remote-plugins.rst
TomasTomecek: good catch!

TomasTomecek:
Updated.

docs/usage/remote-plugins.rst
```diff
@@ -45,3 +46,46 @@ but it can make asynchronous requests, i.e. passing ``async=True``.
 
 You need to run ``:UpdateRemotePlugins`` in Neovim for changes in the specifications to have effect.
 For details see ``:help remote-plugin`` in Neovim.
+
+If you are working on your plugin locally, it may be desirable to run the
+plugin from a git repository directly without the need of installing or
+symlinking it.
```
justinmk: This kind of thing does not need to be spelled out in extreme detail, we have to assume some basic level of familiarity with software development. So we really need to keep the documents concise, to avoid reader fatigue.

docs/usage/remote-plugins.rst
```diff
@@ -45,3 +46,46 @@ but it can make asynchronous requests, i.e. passing ``async=True``.
...
+For local plugin development, it's a good idea to use an isolated vimrc:
+
+.. code-block:: console
+
+    $ cat vimrc
```
justinmk: The `$` makes copy-pasting commands really annoying. Even if it's ambiguous, it's generally not a problem.

docs/usage/remote-plugins.rst
TomasTomecek: I'll be honest, this is something I prefer myself: having documentation as clear as possible. I don't like when I have to read between the lines and figure out details myself the hard way (such as, how do I make nvim pick up my plugin from `$PWD`?).

I completely understand if you don't prefer this style and I have no problem adapting. That's why we do these pull requests and reviews, right?

TomasTomecek:
Thanks Justin! I hope it will help some nvim-plugin-devs newbies (such as myself).

PR #325, opened 2018-04-17
Add lua integration (and update_highlights as an useful example)

bfredl:
Async rplugins (and sync ones for that matter) should conveniently be able to use lua to execute some logic on the nvim side. Rplugins can use `exec_lua(""" ....  _myplugin = {...}""")` or `exec_lua('_myplugin = require("myplugin")')` to define a module with helper functions, and then use `nvim.lua._myplugin.myfunction(..., [async=True])`  to call it.

The `update_highlights` shows an example of the helper function pattern. Though it gets a bit more complex  as we expect multiple copies of the python host, ordinary rplugins need not add the channel_id to the module name like the host does. nvim-ipy will add a more simple example how this can look for an rplugin.

### DIFF
--- a/neovim/api/buffer.py
+++ b/neovim/api/buffer.py
@@ -112,6 +112,29 @@ class Buffer(Remote):
         self.request('nvim_buf_clear_highlight', src_id,
                      line_start, line_end, async_=async_)
 
+    def update_highlights(self, src_id, hls, clear_start=0, clear_end=-1,
+                          clear=False, async_=True):
+        """Add or update highlights in batch to avoid unnecessary redraws.
+
+        A `src_id` must have been allocated prior to use of this function. Use
+        for instance `nvim.new_highlight_source()` to get a src_id for your
+        plugin.
+
+        `hls` should be a list of highlight items. Each item should be a list
+        or tuple on the form `("GroupName", linenr, col_start, col_end)` or
+        `("GroupName", linenr)` to highlight an entire line.
+
+        By default existing highlights are preserved. Specify a line range with
+        clear_start and clear_end to replace highlights in this range. As a
+        shorthand, use clear=True to clear the entire buffer before adding the
+        new highlights.
+        """
+        if clear and clear_start is None:
+            clear_start = 0
+        lua = self._session._get_lua_private()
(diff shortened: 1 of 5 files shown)


neovim/api/nvim.py
```diff
@@ -115,6 +139,12 @@ def _to_nvim(self, obj):
             return ExtType(*obj.code_data)
         return obj
 
+    def _get_lua_helpers(self):
```
justinmk: How about `_get_private_lua` or just `_get_lua`. And doc like this:

```
"""
Returns a wrapper to a private Lua module associated with this python client, providing these functions:
...
"""
```

The word "helpers" needs to go away in all of our projects, it is useless and seems to proliferate as soon as it's used once.

neovim/api/buffer.py
```diff
@@ -110,6 +110,23 @@ def clear_highlight(self, src_id, line_start=0, line_end=-1, async=True):
...
+        hls should be a list of highlight items. Each item should be a list or tuple on the form
+        ("GroupName", linenr, col_start, col_end) or ("GroupName", linenr) to highlight an entire line.
+
+        By default existing highlights is preserved. Specify a line range with clear_start and clear_end
+        to replace highlights in this range. As a shorthand, use clear=True to clear the entire
+        buffer before adding the new highlights.
+        """
+        if clear and clear_start is None:
+            clear_start = 0
+        helpers = self._session._get_lua_helpers()
```
justinmk: lua = self._session._get_lua()

bfredl:
Now without `helpers`. I will add tests later.

bfredl:
Updated with tests. Also use only `async_` for new functions.

justinmk:
LGTM. Will the docs autogenerate? I think it's cool enough to be mentioned in the README.md too.

bfredl:
readthedocs has a "master" channel that should update on push to master. I plan to do a release very soon after this, so plugins can prepare for 3.7 (and use this), the release after that (coordinated with nvim 0.3?) will do the formal rename.

bfredl:
Merged. I will update docs for next release in separate PR.

PR #328, opened 2018-04-26
Add Buffer.__repr__ and Window.__repr__

blueyed:


justinmk:
somewhat related, "lgtm" service also noticed some other things: https://lgtm.com/projects/g/neovim/python-client/alerts/

bfredl:
As an alternative, could add a generic `__repr__` in the `Remote` base class and cover all three objects.

blueyed:
@bfredl 

> As an alternative, could add a generic __repr__ in the Remote base class and cover all three objects.

They have different properties though.

I've implemented it now for all three.

bfredl:
@blueyed They all have the same _essential_ property: the object handle. Also I'm not sure we want `__repr__` to use sync requests: one  should be able to print object representations in a debug statement, even at a place where sync requests are not safe.

bfredl:
Note that most other attributes than the handle requires sync requests. What I had in mind was something like
```    def __repr__(self):
       """Get text representation of the tabpage."""
       return '<pynvim.{}(handle={})>'.format(self.__class__.__name__,
                                             self.handle)

```

which can be in the `Remote` class.

blueyed:
Added a commit to remove remote calls altogether.

neovim/api/buffer.py
```diff
@@ -26,6 +26,13 @@ class Buffer(Remote):
 
     _api_prefix = "nvim_buf_"
 
+    def __repr__(self):
+        """Get text representation of the buffer."""
+        return 'Buffer(number=%r, handle=%r)' % (
+            self.number,
```
bfredl: `number` and `handle` are synonymous. After this, we can just as well have a shared implementation in `Remote`.

justinmk:
LGTM, but tests are failing.

```
file /home/travis/build/neovim/python-client/test/test_buffer.py, line 6
  def test_repr(request, vim):
file /home/travis/build/neovim/python-client/test/conftest.py, line 70
  @pytest.fixture
  def no_remote_calls(vim, mocker):
E       fixture 'mocker' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cleanup_func, doctest_namespace, monkeypatch, no_remote_calls, pytestconfig, record_property, record_xml_attribute, record_xml_property, recwarn, tmpdir, tmpdir_factory, vim, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.
```

blueyed:
Do you think the `no_remote_calls` fixture is worth the pytest-mock dependency?
If so I'll fix the build/installation.
/cc @bfredl 
Also note that I haven't included the suggested `pynvim.` prefix, which is not typical with `__repr__` methods AFAIK.

bfredl:
Lets skip the mock test for now, it seems overcomplicated for only this functionality.

The output is probably fine as it is. The reason for the `<' and `>` would say that it is non-reconstructible (unlike `repr(json_like_object)` ) but we can skip that, it should be understood anyway.

blueyed:
Amended, and added the `<…>`.

test/test_buffer.py
```diff
@@ -3,6 +3,10 @@
 from neovim.compat import IS_PYTHON3
 
 
+def test_repr(vim):
+    assert repr(vim.current.buffer) == "<Buffer(handle=2)>"
```
bfredl: hmm, looks a bit fragile, but test run order seems deterministic. So guess we can keep this for now and change it to a regex later if it turns out to be too much churn.

Issue #334, opened 2018-04-28
Unknown function


meijieru:
here are the rplugin

```python
import neovim


@neovim.plugin
class Main(object):

    def __init__(self, vim):
        self.vim = vim

    @neovim.function('test_python')
    def test_python(self, args):
        self.vim.command('echo "hello from TestPython"')

    @neovim.function('TestPython')
    def test_python_1(self, args):
        self.vim.command('echo "hello from TestPython"')
```

put this script in `~/.config/nvim/rplugin/python3` and run `UpdateRemotePlugins`, generate `rplugin.vim` in `.local/share/nvim`

```vim
" node plugins
" python3 plugins
call remote#host#RegisterPlugin('python3', '/home/jrmei/workspace/other/dotfiles/dotfiles/vim/nvim/rplugin/python3/auxlib.py', [
      \ {'sync': v:false, 'name': 'test_python', 'type': 'function', 'opts': {}},
      \ {'sync': v:false, 'name': 'TestPython', 'type': 'function', 'opts': {}},
     \ ])
" ruby plugins
" python plugins
```

now if run `call TestPython()`, if will output correctly. But if run `call test_python()`, it complains unknown function.

bfredl:
Plugin defined functions and commands must begin with uppercase. But we should throw an error for this as early as possible, ideally by nvim core so it covers hosts in all languages.

meijieru:
Yeah, that would be ideal

meijieru:
Additional question.

Here are vimscript part of my plugin
```vim
call AuxlibEnableLogging()
```
However, vim also complains `E117: Unknown function: AuxlibEnableLogging`.
But after vim started, directly run `call AuxlibEnableLogging()` does not trigger any error.

justinmk:
rplugin decorators will be deprecated by https://github.com/neovim/pynvim/issues/567

PR #335, opened 2018-04-28
[WIP] Use function name for decorators

stsewd:
Sometimes is kind of redundant to write the function/command/autocmd name on each decorator. Also, this can help with cases like #334.

Please, let me know if you want this feature so, I can continue :), thanks

What is missing?

- [ ] Update docs

### DIFF
--- a/neovim/plugin/decorators.py
+++ b/neovim/plugin/decorators.py
@@ -42,11 +42,16 @@ def rpc_export(rpc_method_name, sync=False):
     return dec
 
 
-def command(name, nargs=0, complete=None, range=None, count=None, bang=False,
-            register=False, sync=False, allow_nested=False, eval=None):
+def command(name=None, nargs=0, complete=None, range=None,
+            count=None, bang=False, register=False, sync=False,
+            allow_nested=False, eval=None):
     """Tag a function or plugin method as a Nvim command handler."""
     def dec(f):
-        f._nvim_rpc_method_name = 'command:{}'.format(name)
+        if not name:
+            command_name = capitalize_name(f.__name__)
+        else:
+            command_name = name
+        f._nvim_rpc_method_name = 'command:{}'.format(command_name)
         f._nvim_rpc_sync = sync
         f._nvim_bind = True
         f._nvim_prefix_plugin_path = True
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -8,7 +8,7 @@ deps=
   pytest
   pytest-xdist
   pyuv: pyuv
-commands=python -m pytest -s
+commands=python -m pytest -s {posargs}
```
stsewd: I'm adding this so I can do `tox -e py36 test/test_decorators.py`

neovim/plugin/decorators.py
```diff
@@ -141,14 +155,23 @@ def dec(f):
...
-            'name': name,
+            'name': function_name,
             'sync': rpc_sync,
             'opts': opts
         }
         return f
     return dec
 
 
+def capitalize_name(name):
```
bfredl: I do not like this kind of magic. For plugins that use this, it would be better to break python conventions and name the implementation function literally what the vimL side name is. Also, `_` is not forbidden in vimL.

neovim/plugin/decorators.py
```diff
@@ -80,18 +85,22 @@ def dec(f):
...
+            'name': command_name,
             'sync': rpc_sync,
             'opts': opts
         }
         return f
     return dec
 
 
-def autocmd(name, pattern='*', sync=False, allow_nested=False, eval=None):
+def autocmd(name=None, pattern='*', sync=False, allow_nested=False, eval=None):
```
bfredl: It can be a concern that a plugin potentially can have multiple implementations of the autocmd with disjoint patterns. At least there should be a warning in the docs. (python doesn't give a warning on method shadowing, though we could hack it with a metaclass).

stsewd: I wasn't sure about the autocmd, it doesn't have much sense here I guess.

stsewd: Maybe the parameter should be named `event`, that's why I was confused here.

neovim/plugin/decorators.py
stsewd: Well, a matter of style I think. I would no like to have a function/command named `My_function_something`, and you can still pass a name if people want that name.

justinmk:
https://github.com/neovim/neovim/issues/27949 will greatly simplify the implementation and interface of remote plugins, and will remove the need for decorators (which are just sugar to do things that can/should be done in the plain Lua part of a remote plugin).

PR #340, opened 2018-06-08
[WIP] buffer: support attachment in hosted plugins

bfredl:
Warning: very rough sketch, but the intention is that multiple plugins can register buffer notifications on the same or different buffers without stepping on each other's toe.

At least the following works: (script for quick test, but does work the same in rplugin)
 
```python
import vim
b = vim.current.buffer
def cb(buf, changedtick, lines):
     global l
     l = lines

b.attach(cb)
```

Not sure how to best represent the update, plugins will likely want to use the full new buffer state, but likely they also want to know `start, old_end, new_end`


neovim/api/buffer.py
```diff
@@ -155,6 +154,32 @@ def number(self):
...
+            self.api.attach(True, {})
+        else:
+            a = sess.attached_buffers[self.handle]
+            a.callbacks.append(cb)
+            cb(self, a.changedtick, a.lines)
+
+        def detach():
+            for i in range(len(a.callbacks)):
+                if a.callbacks[i] is cb:
+                    del a.callbacks[i]
```
bfredl: NB: should detach when callbacks goes to zero.

pierreprinetti:
Hi :wave: 
What's the current status of this work? What is left to do before merging (on top of a rebase)?

PR #344, opened 2018-07-14
Fix PendingDeprecationWarning with Packer.encoding

blueyed:
Fixes:

  …/neovim-python-client/neovim/msgpack_rpc/msgpack_stream.py:24: PendingDeprecationWarning: encoding is deprecated.
    unicode_errors=unicode_errors_default)

Also configures pytest to display these warnings.

### DIFF
--- a/neovim/msgpack_rpc/msgpack_stream.py
+++ b/neovim/msgpack_rpc/msgpack_stream.py
@@ -20,8 +20,7 @@ class MsgpackStream(object):
     def __init__(self, event_loop):
         """Wrap `event_loop` on a msgpack-aware interface."""
         self.loop = event_loop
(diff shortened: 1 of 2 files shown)


bfredl:
Thanks!

bfredl:
Unfortunately this seems to break some configurations (such as mine, though I have latest `msgpack` package), so reverted.

blueyed:
Do you get the warnings then when running the tests yourself (with the pytest settings from this PR)?

bfredl:
Looking at the source code `encoding=` does in fact not default to `"utf-8"` on py3, rather `None`, what the py api function used does with None I don't know, but I suppose it doesn't mean UTF-8 all of the time.

PR #345, opened 2018-07-14
Use nvim_ui_attach/nvim_ui_detach

blueyed:
`ui_attach` and `ui_detach` are deprecated.

This handles the old signature, and issues a DeprecationWarning itself
then.

### DIFF
--- a/neovim/api/nvim.py
+++ b/neovim/api/nvim.py
@@ -1,6 +1,7 @@
 """Main Nvim interface."""
 import os
 import sys
+import warnings
 from functools import partial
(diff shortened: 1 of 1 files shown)


bfredl:
Why not use kwargs for options? This is what python-gui does, and needs no deprecation for `rgb`, just allow it as third pos arg as always.

blueyed:
Ok, makes sense. Fixed.

blueyed:
Please squash-merge then.

bfredl:
thanks, merged

PR #346, opened 2018-07-14
[RFC] Add RemoteApi.__dir__ with info from nvim_get_api_info

blueyed:
This can be used by IPython's completion for example.

Is the selection sane?
Could there be more?
AFAICS only functions are supported/used there.

### DIFF
--- a/neovim/api/common.py
+++ b/neovim/api/common.py
@@ -60,9 +60,25 @@ class RemoteApi(object):
         self._obj = obj
         self._api_prefix = api_prefix
 
+        self._api_names = None
+
     def __getattr__(self, name):
         """Return wrapper to named api method."""
         return functools.partial(self._obj.request, self._api_prefix + name)
+
+    def __dir__(self):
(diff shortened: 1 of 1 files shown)


neovim/api/common.py
```diff
@@ -60,10 +60,26 @@ def __init__(self, obj, api_prefix):
...
         """Return wrapper to named api method."""
         return functools.partial(self._obj.request, self._api_prefix + name)
 
+    def __dir__(self):
+        """Return info via nvim_get_api_info.
+
+        This can be used for introspection, and especially completion.
+        """
+        if self._api_names is None:
+            api_info = self._obj.request('nvim_get_api_info')
```
bfredl: there is a shared metadata object as `nvim.metadata` though it is a bit tricky to access it as `_obj` could be either a `Nvim` or a `Remote`. Guess you could do something like `nvim = _obj._session if isinstance(obj,Remote) else _obj`

PR #350, opened 2018-07-17
host: add client info

bfredl:


neovim/plugin/host.py
```diff
@@ -141,6 +144,15 @@ def _load(self, plugins):
                 error(err)
                 self._load_errors[path] = err
 
+        if len(plugins) == 1 and has_script:
+            kind = "script"
+        else:
+            kind = "rplugin"
+        name = "python{}-{}-host".format(sys.version_info[0], kind)
+        methods = { "poll": {}, "specs": {"nargs":1}, "shutdown": {} }
```
justinmk: is it possible to write a test that checks that this has all of the pynvim methods?

bfredl: added the test.

PR #358, opened 2018-11-07
improve message for out-of-thread requests

bfredl:


bfredl:
flake8 seems to randomly enable different style checks with new version, fixing them later...

PR #359, opened 2018-11-17
Rename occurrences of 'neovim' to 'pynvim'

mhinz:


mhinz:
I guess this needs to be changed as well, right?

https://github.com/neovim/python-client/blob/e1e04e1a86b614a08b7b725e27d8f23e85f25c69/.travis.yml#L8

bfredl:
thanks!

PR #356, opened 2018-10-08
add python3.7 to travis

bfredl:

PR #360, opened 2018-11-18
Travis: test with Python 3.7

blueyed:
Closes https://github.com/neovim/python-client/pull/356.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+dist: xenial
 language: python
 env:
   global:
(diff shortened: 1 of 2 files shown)


bfredl:
Seems to not work with pypy, could it stay on 14.04?

blueyed:
Yes, done.

.travis.yml
```diff
@@ -9,14 +9,17 @@ matrix:
   include:
     - python: 3.6
       env: CI_TARGET=checkqa TOXENV=checkqa
+    - python: pypy
```
bfredl: could this be moved abouve `checkqa`?  Would be nice to keep all test suite runs grouped together, and the lint stuff at the bottom.

bfredl:
I did the ordering change myself and then merged this, thanks!

PR #361, opened 2018-11-18
Fix flake8 config

blueyed:
Ignore W504 (using W503):

> W504 line break after binary operator


bfredl:
Nice!

justinmk:
Why don't we move the operators to the start of the next line instead? That's the style used in nvim core, and now also recommended by pep8.

blueyed:
@justinmk 
I prefer that also, but changed the config according to the current style.
Feel free to change it.

PR #363, opened 2018-11-18
scripts/logging_statement_modifier.py: use print()  [ci skip]

blueyed:


justinmk:
note also https://github.com/neovim/pynvim/issues/250 describes improvements to the logging that we still haven't implemented ...

PR #364, opened 2018-11-18
Avoid deprecated symbols

### LINKED ISSUE #351: Deprecated import
System: MacOS
Neovim: **0.3.0**
Python: **3.7**
Python client (neovim): **0.2.6**

> DeprecationWarning: the **imp** module is deprecated in favour of **importlib**; see the module's documentation for alternative uses

It is triggered by the plugin `vim-visual-multi` when using the `python3` api.

justinmk:
- "imp" was deprecated in python 3.7. closes #351
- "encoding" was deprecated in msgpack module.
    - `raw_as_bytes` was added in 0.5.2 and is recommended going forward, but it's not clear to me if we have any reason to specify it. Advice welcome.
    - ref msgpack/msgpack-python@5534d0c#diff-88b99bb28683bd5b7e3a204826ead112R40


pynvim/plugin/script_host.py
```diff
@@ -38,7 +40,10 @@ def __init__(self, nvim):
         """Initialize the legacy python-vim environment."""
         self.setup(nvim)
         # context where all code will run
-        self.module = imp.new_module('__main__')
+        if IS_PYTHON3_7:
+            self.module = importlib.util.module_from_spec('__main__')
+        else:
+            self.module = imp.new_module('__main__')
```
justinmk: Hope I didn't do something stupid here.

README.md
```diff
@@ -56,13 +56,23 @@ See the [Python Plugin API](http://pynvim.readthedocs.io/en/latest/usage/python-
...
-If you change the code, you need to run
-```sh
-pip2 install .
-pip3 install .
-```
-for the changes to have effect. For instructions of testing and troubleshooting,
-see the [development](http://pynvim.readthedocs.io/en/latest/development.html) documentation.
+Use (and activate) a local virtualenv.
+
+    python -m virtualenv --python=python3 venv
```
bfredl: Is `venv` the name of a specifiic virtualenv? A bit unfortunate as `venv` is a basic virtualenv implementation shipped with python.

blueyed: That's the path.
I would rather use `.venv` instead, and the `venv` module (with py3): `python -m venv .venv`

justinmk: @blueyed Thanks. Is there a way to specify different versions with `python -m venv`? 

```
$ python3 -m venv --python=python3.7 py37                                                                                                                   
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear]      
            [--upgrade] [--without-pip] [--prompt PROMPT]        
            ENV_DIR [ENV_DIR ...]                                              
venv: error: unrecognized arguments: --python=python3.7                          
```

blueyed: It uses (only) the existing one (the caller) I assume.

PR #372, opened 2018-12-03
host: catch and show nvim_error_event

bfredl:


justinmk:
Is it testable by calling an invalid method?

bfredl:
Yes, though tests for host behavior is currently in `neovim/neovim` repo. By using this as library, as the test suite here tests, nothing is changed, `nvim_error_event` is just like any other notification.

PR #377, opened 2018-12-28
rename stdout fd in stdio mode

### LINKED ISSUE #60: Trouble with subprocess stdio
The following crashes the python host: (which is an improvment, som time ago this crashed nvim :p )

```
python from subprocess import call
python call(["echo","S"])
```

This because subprocesses inherit stdout and will inject output directly into the binary msgpack stream. stdout was written to by a helper process inside a library, so at first I couldn't understand why on earth nvim complained about a malformatted message that the client obviously never sent. One could argue that a correcty behaving plugin/library should redirect stdio itself (`subprocess` makes this easy), but the host should error out a better way than just crashing.

bfredl:
Fixes #60 and https://github.com/neovim/neovim/issues/9397.

Or at least it avoids the crash. We should also make the stderr visible i e when `UpdateRemotePlugins`, but that is nvim core issue. (technically each rplugin host could do it using an internal pipe, but we can just as well have a one-size-fits-all solution in nvim core).

### DIFF
--- a/pynvim/msgpack_rpc/event_loop/asyncio.py
+++ b/pynvim/msgpack_rpc/event_loop/asyncio.py
@@ -104,10 +104,15 @@ class AsyncioEventLoop(BaseEventLoop, asyncio.Protocol,
         self._loop.run_until_complete(coroutine)
         debug("native stdin connection successful")
 
+        # Make sure subprocesses don't clobber stdout,
+        # send the output to stderr instead.
+        rename_stdout = os.dup(sys.stdout.fileno())
+        os.dup2(sys.stderr.fileno(), sys.stdout.fileno())
+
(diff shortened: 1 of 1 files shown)


pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -104,10 +104,14 @@ def _connect_stdio(self):
         self._loop.run_until_complete(coroutine)
         debug("native stdin connection successful")
 
+        stdout_fd = sys.stdout.fileno()
+        rename_stdout = os.dup(stdout_fd)
```
justinmk: is `stdout_fd` needed?

bfredl:
Tested on Linux and Win10.

PR #389, opened 2019-04-01
doc, tests: specify --headless

justinmk:


test/test_vim.py
```diff
@@ -90,9 +90,9 @@ def test_vars(vim):
 
 
 def test_options(vim):
-    assert vim.options['listchars'] == 'tab:> ,trail:-,nbsp:+'
-    vim.options['listchars'] = 'tab:xy'
-    assert vim.options['listchars'] == 'tab:xy'
+    assert vim.windows[0].options['listchars'] == 'tab:> ,trail:-,nbsp:+'
+    vim.windows[0].options['listchars'] = 'tab:xy'
+    assert vim.windows[0].options['listchars'] == 'tab:xy'
```
justinmk: Not sure why this test only _now_ started failing, but anyways this matchs `nvim_get_option` behavior. I guess it got more strict in the last few months but I somehow don't remember that change.

bfredl: @justinmk because `nvim_win_get_option` is buggy. I should finish the PR that fixes that...

bfredl: Though probably that issue was irrelevant, as `set_option` should still match it..

Issue #391, opened 2019-05-02
Startup script for Ipython 


mrsalo:
In order to use and test out pynvim it would be beneficial to have a startup script for IPython that populates a connection to a local NeoVim instance without further setup required. 
At least in  Arch Linux AFAIK every NeoVim instance populates its API endpoint under /tmp/nvimRANDOM/0. This makes it possible to connect to every instance, echo a identifying message to the user to enable him/her choosing to which instance to connect to. If there is only one instance it is also possible to connect to it directly. 
Example Usage:
![Peek 2019-05-02 22-06](https://user-images.githubusercontent.com/3208694/57103711-d3234680-6d26-11e9-8aaa-11a398586ffd.gif)

justinmk:
Sure. Here's something similar in my config: https://github.com/justinmk/config/blob/ad5b792049b352274d4cbd3525a2aff6ce296a7e/.config/nvim/init.vim#L1339-L1361

I think it might be a good idea to have some sort of convention where each client implements something like this. Then the user can just `rpcrequest({chan}, 'repl')`.

Or it could simply be a plugin included with Nvim core runtime.

mrsalo:
I thought about staying completely in neovim by starting Ipython in conjunction with `:terminal` similar as you did in your init.vim, too. The only problem is that i cannot use the convenient `nvim.current .buffer` anymore as i am referring to the `:terminal` buffer then.
That's why I chose to start IPython in another terminal. 
I think I do not fully understand your `rpcrequest({chan}, 'repl')` suggestion, could you elaborate on this?

PR #392, opened 2019-05-03
Startup script repl

mrsalo:
#391

### DIFF
--- a/scripts/nvim_startup.py
+++ b/scripts/nvim_startup.py
@@ -1,0 +1,65 @@
+""" Startup script for IPython that will connect to a running NeoVim instance
+automatically or ask the user which NeoVim instance to connect to.
+After connection there is a variable called "nvim" that gives acccess to the
+endpoint.
+"""
+import atexit
+from glob import glob
+import pynvim
+
+
+class NVimAutoAttachException(Exception):
+    '''Exception class that will be raised by this script'''
+
+
+def __let_user_choose_instance(nvim_instances):
+    for idx, inst in enumerate(nvim_instances):
+        with pynvim.attach('socket', path=inst) as nvim_endpoint:
+            nvim_endpoint.command(
+                f'echon "I am Instance Nr {idx}: " $NVIM_LISTEN_ADDRESS'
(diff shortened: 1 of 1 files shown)


justinmk:
what is step-by-step the workflow for using this? That should be mentioned in the README.

I would think invoking something from Nvim would be more useful, rather than making the user choose from a list.

mrsalo:
I updated the README. I don't know if this is already enough documentation. 
I would also prefer the invocation through nvim, but it has one flaw:
When I use the `:term` feature to start a terminal and then e.g. IPython I am perfectly able to connect to the current instance. But the current buffer (the one i get using `nvim.current.buffer`) is the terminal buffer wher IPython runs and not the one I wanted to modify through the REPL initially. 
Hope this gif makes it a little bit clearer:
Peek 2019-05-03 14-06

justinmk:
Actually, I don't think it makes sense to put this in pynvim repo. Most pynvim users won't have the `scripts/` AFAIK. So if anything this should be a plugin or maybe `contrib/` in the neovim repo.

scripts/nvim_startup.py
```diff
@@ -0,0 +1,62 @@
...
+NeoVim instance automatically or ask the user which NeoVim instance to connect 
+to. After connection there is a variable called "nvim" that gives acccess to 
+the endpoint.
+"""
+import atexit
+from glob import glob
+import pynvim
+
+
+class NVimAutoAttachException(Exception):
```
justinmk: "Nvim", not "NVim"

mrsalo:
Ok, what would you think of putting it in the pynvim module or in a submodule and not as a startup script but as a function?
If you think it's not worth the hassle that`s also fine. Just found it easier to use than the standard procedure.

justinmk:
> Ok, what would you think of putting it in the pynvim module or in a submodule and not as a startup script but as a function?

That was my first thought. I guess it makes sense because then Nvim core only needs to look for that function. The function can be mentioned by `nvim_set_client_info`. I think `open_repl` might be a good name.

PR #397, opened 2019-05-15
tests: global and local options

mhinz:
While `'listchars'` is a global option in Nvim 0.3.5, in HEAD it was changed to be local to the window.

### DIFF
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -90,9 +90,15 @@ def test_vars(vim):
 
 
 def test_options(vim):
-    assert vim.windows[0].options['listchars'] == 'tab:> ,trail:-,nbsp:+'
-    vim.windows[0].options['listchars'] = 'tab:xy'
-    assert vim.windows[0].options['listchars'] == 'tab:xy'
+    assert vim.options['background'] == 'dark'
+    vim.options['background'] = 'light'
+    assert vim.options['background'] == 'light'
+
+
+def test_local_options(vim):
+    assert vim.windows[0].options['foldmethod'] == 'manual'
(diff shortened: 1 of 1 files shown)


test/test_vim.py
```diff
@@ -90,9 +90,15 @@ def test_vars(vim):
 
 
 def test_options(vim):
-    assert vim.windows[0].options['listchars'] == 'tab:> ,trail:-,nbsp:+'
-    vim.windows[0].options['listchars'] = 'tab:xy'
-    assert vim.windows[0].options['listchars'] == 'tab:xy'
+    assert vim.options['listchars'] == 'tab:> ,trail:-,nbsp:+'
```
bfredl: but this didn't work on master right? In the meanwhile we should use some other global option.

mhinz: But that way it should work with `'listchars'` being either global _or_ locally?

mhinz: Oh, it does not. Nevermind then. I'll use another option.

bfredl: Hmm are you sure? Was this test not changed recently because it did fail with nvim master?

justinmk: > Hmm are you sure? Was this test not changed recently because it did fail with nvim master?

yes https://github.com/neovim/pynvim/commit/5a329f221123efbb5a5515692260eb87469ebd39

bfredl: `vim.options['listchars']` doesn't work for me in manual testing on nvim master. (One can argue it _should_ work, but it doesn't).

mhinz: No, sorry, that was my bad. `vim.options['listchars']` wouldn't work with HEAD.

PR #405, opened 2019-08-19
[RDY] Add support for proper deletion, raising KeyError where appropriate

blueyed:
Fixes https://github.com/neovim/pynvim/issues/402

TODO:

- [x] rebase after https://github.com/neovim/pynvim/pull/408
- [x] squash, rebase after https://github.com/neovim/pynvim/pull/412

### DIFF
--- a/pynvim/api/common.py
+++ b/pynvim/api/common.py
@@ -6,6 +6,10 @@ from msgpack import unpackb
 from ..compat import unicode_errors_default
 
 
+class NvimError(Exception):
+    pass
+
+
 class Remote(object):
 
     """Base class for Nvim objects(buffer/window/tabpage).
@@ -26,7 +30,8 @@ class Remote(object):
         self.handle = unpackb(code_data[1])
         self.api = RemoteApi(self, self._api_prefix)
         self.vars = RemoteMap(self, self._api_prefix + 'get_var',
-                              self._api_prefix + 'set_var')
+                              self._api_prefix + 'set_var',
+                              self._api_prefix + 'del_var')
         self.options = RemoteMap(self, self._api_prefix + 'get_option',
                                  self._api_prefix + 'set_option')
 
@@ -65,8 +70,16 @@ class RemoteApi(object):
         return functools.partial(self._obj.request, self._api_prefix + name)
 
 
-class RemoteMap(object):
+def transform_keyerror(exc):
+    if isinstance(exc, NvimError):
+        if exc.args[0].startswith('Key not found:'):
+            return KeyError(exc.args[0])
+        if exc.args[0].startswith('Invalid option name:'):
+            return KeyError(exc.args[0])
+    return exc
 
+
+class RemoteMap(object):
     """Represents a string->object map stored in Nvim.
 
     This is the dict counterpart to the `RemoteSequence` class, but it is used
@@ -75,17 +88,23 @@ class RemoteMap(object):
 
     It is used to provide a dict-like API to vim variables and options.
     """
+    _set = None
+    _del = None
 
-    def __init__(self, obj, get_method, set_method=None):
+    def __init__(self, obj, get_method, set_method=None, del_method=None):
         """Initialize a RemoteMap with session, getter/setter."""
         self._get = functools.partial(obj.request, get_method)
-        self._set = None
         if set_method:
             self._set = functools.partial(obj.request, set_method)
+        if del_method:
+            self._del = functools.partial(obj.request, del_method)
 
     def __getitem__(self, key):
         """Return a map value by key."""
-        return self._get(key)
+        try:
+            return self._get(key)
+        except NvimError as exc:
+            raise transform_keyerror(exc)
 
     def __setitem__(self, key, value):
         """Set a map value by key(if the setter was provided)."""
@@ -95,9 +114,12 @@ class RemoteMap(object):
 
     def __delitem__(self, key):
         """Delete a map value by associating None with the key."""
-        if not self._set:
+        if not self._del:
             raise TypeError('This dict is read-only')
-        return self._set(key, None)
+        try:
+            return self._del(key)
+        except NvimError as exc:
+            raise transform_keyerror(exc)
 
     def __contains__(self, key):
         """Check if key is present in the map."""
@@ -110,8 +132,8 @@ class RemoteMap(object):
     def get(self, key, default=None):
         """Return value for key if present, else a default value."""
         try:
-            return self._get(key)
-        except Exception:
+            return self.__getitem__(key)
+        except KeyError:
             return default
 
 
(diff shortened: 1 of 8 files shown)


pynvim/api/nvim.py
```diff
@@ -579,3 +579,7 @@ def __call__(self, *args, **kwargs):
 
 class NvimError(Exception):
     pass
+
+
+class NvimKeyError(KeyError, NvimError):
+    pass
```
blueyed: Should we move errors to `nvim.api.errors` or similar?  This would likely aboid circular imports then (as mentiones in `transform_keyerror`).

pynvim/util.py
```diff
@@ -40,3 +41,22 @@ def get_client_info(kind, type_, method_spec):
 
 
 VERSION = Version(major=0, minor=3, patch=2, prerelease='')
+
+
+def transform_keyerror(f):
+    """Decorator to transform "Key not found" NvimErrors into KeyErrors."""
```
blueyed: Also handles "Invalid option name" by now.
I.e. doc should be updated / generalized.

Might make sense also to have different wrappers for vars and options (for clarity).

pynvim/api/nvim.py
```diff
@@ -76,7 +76,7 @@ def from_session(cls, session):
         queries Nvim metadata for type information and sets a SessionHook for
         creating specialized objects from Nvim remote handles.
         """
-        session.error_wrapper = lambda e: NvimError(e[1])
+        session.error_wrapper = lambda e: NvimError(e[1].decode())
```
blueyed: This is done to have strings with exception arguments, instead of bytes (py3).

pynvim/util.py
```diff
@@ -40,3 +41,22 @@ def get_client_info(kind, type_, method_spec):
...
+    def wrapper(*args, **kwargs):
+        # XXX: # not on top, because of circular imports.
+        from .api.nvim import NvimError, NvimKeyError
+
+        try:
+            return f(*args, **kwargs)
+        except Exception as exc:
+            if isinstance(exc, NvimError):
+                if exc.args[0].startswith('Key not found:'):
+                    raise NvimKeyError(exc.args[0])
```
bfredl: Why not just use `KeyError` ? `NvimKeyError` sounds like typical overengineering.

pynvim/util.py
```diff
@@ -40,3 +41,22 @@ def get_client_info(kind, type_, method_spec):
...
 
 VERSION = Version(major=0, minor=3, patch=2, prerelease='')
+
+
+def transform_keyerror(f):
+    """Decorator to transform "Key not found" NvimErrors into KeyErrors."""
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        # XXX: # not on top, because of circular imports.
+        from .api.nvim import NvimError, NvimKeyError
```
bfredl: Why not have this in the same module as `RemoteMap` where it is used, it should avoid the import issue? Also not sure this is made simpler/clearer by using a decorator, we could just write out `try/expect` in both `__getitem__` and `__delitem__` (and call a value-changing helper function).

pynvim/util.py
blueyed: There is a test that expects to see `vim.error`.
And it makes sense to have it for backward compatibility I think.

pynvim/util.py
blueyed: the import issue is for `NvimError`.
But it could be moved there maybe.

> and call a value-changing helper function

What do you mean?

Yes, using a decorator is not really necessary.

pynvim/util.py
bfredl: hmm, so it is required for vim compat then.

bfredl: Or is it? `vim.current.buffer.vars` and `.option` seem to return `KeyError` only.

blueyed: I've meant compat with previous pynvim versions.
Our test for this: https://github.com/blueyed/python-client/blob/b9f0061c6d862062aa4596478029ddf70472809d/test/test_buffer.py#L156-L161

pynvim/util.py
bfredl: I mean we could still factor out the check of the error value to a helper function (i e check if the string matches, and return a different error object).

pynvim/util.py
bfredl: `vim.error` is a vim-compat feature, no need to be compatible with a vim behavior which does not exist. The question is if some plugin is using `NvimError` explicitly to check for missing key which I wouldn't expect (the canonical way if you expect it to be missing is `.get()` ), perhaps checking the matches on github is enough.

blueyed: Could not find usages via https://github.com/search?p=2&q=except+nvimerror+vim+vars+extension%3Apy&type=Code - changed it to use KeyError directly.

pynvim/util.py
blueyed: ok, done.

pynvim/api/common.py
```diff
@@ -26,7 +30,8 @@ def __init__(self, session, code_data):
         self.handle = unpackb(code_data[1])
         self.api = RemoteApi(self, self._api_prefix)
         self.vars = RemoteMap(self, self._api_prefix + 'get_var',
-                              self._api_prefix + 'set_var')
+                              self._api_prefix + 'set_var',
+                              self._api_prefix + 'del_var')
         self.options = RemoteMap(self, self._api_prefix + 'get_option',
                                  self._api_prefix + 'set_option')
```
blueyed: Should we pass an argument to `RemoteMap` to only handle specific errors as key errors?
(I mean `self.vars` should never see 'Invalid option name:')

pynvim/api/nvim.py
blueyed: Should I take this out of here for now already?

Or can we finish it via this PR?

bfredl: Should use `decode("utf-8", "surrogateescape")` on python3 and no decode on python2 ( might be some function to reuse that does this)

blueyed: > no decode on python2

Why not?  Do we explicitly want no unicode string?

blueyed: There's `decode_if_bytes`.  Taken out into https://github.com/neovim/pynvim/pull/412.

bfredl: We are explicitly using the standard `str` type in both versions per default. Trying to "improve" string semantics on py2 is a non-goal at this point.

pynvim/api/common.py
bfredl: We probably don't need to worry about it.

blueyed:
Thanks for the approvals, please review https://github.com/neovim/pynvim/pull/412 also.
I'll then rebase and merge this one here after.

PR #408, opened 2019-08-19
ci: fix coverage reporting

blueyed:
Add "cov" tox factor, which should get used by tox-travis then.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project: true
+    patch: true
+    changes: true
+comment: false
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+source = .
+branch = true
+
+[report]
+show_missing = true
+include = pynvim/*,test/*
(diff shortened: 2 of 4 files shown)


blueyed:
Might not need codecov - initially fixed it because Scrutinizer was failing/erroring due to it.

blueyed:
It was uploaded for Scrutinizer (https://travis-ci.org/neovim/pynvim/jobs/573872712#L526), but it is still pending: https://scrutinizer-ci.com/g/neovim/pynvim/inspections/f60fbc9e-04de-4810-8196-7520f3b83117

Apparently waiting for the upload for master, which is not going to happen.

Therefore I think this is ready, and then should be fixed for future PRs.

blueyed:
@bfredl @justinmk 
OK to merge?

bfredl:
looks ok to me.

justinmk:
I don't think we want those codecov comments on PRs. Does this enable those somehow?

blueyed:
@justinmk 
They are disabled in the config: https://github.com/neovim/pynvim/pull/408/files#diff-1f7b4b85ed8525c5239f741431a72872R6
Probably not working properly for the initial addition of it etc.
And while it might still happen due to bugs with them (as seen with neovim/neovim sometimes), they should not appear normally.

PR #410, opened 2019-08-20
 Raise KeyError for missing keys

noelevans:
Fixing issue neovim#402
"Raise KeyError for missing options / vars".

### DIFF
--- a/pynvim/api/common.py
+++ b/pynvim/api/common.py
@@ -85,7 +85,10 @@ class RemoteMap(object):
 
     def __getitem__(self, key):
         """Return a map value by key."""
-        return self._get(key)
+        try:
+            return self._get(key)
+        except Exception:
+            raise KeyError
 
     def __setitem__(self, key, value):
         """Set a map value by key(if the setter was provided)."""
--- a/pynvim/api/common.py
+++ b/pynvim/api/common.py
@@ -88,7 +88,7 @@ class RemoteMap(object):
         try:
             return self._get(key)
         except Exception:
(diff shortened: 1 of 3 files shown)


pynvim/api/common.py
```diff
@@ -85,7 +85,10 @@ def __init__(self, obj, get_method, set_method=None):
 
     def __getitem__(self, key):
         """Return a map value by key."""
-        return self._get(key)
+        try:
+            return self._get(key)
+        except Exception:
+            raise KeyError
```
mhinz: Is the generic `KeyError` the proper thing to throw for the API?

Maybe throw `NvimAPIKeyError` which derives from `KeyError`? Would be more precise and `exception KeyError` in user code would still work.

bfredl:
> Maybe throw NvimAPIKeyError which derives from KeyError? 

We do not need to create a new type for every specific place in which a specific error can be thrown, this is just typical overengineering. You can already check the error for vars/options only just by using `try/except` around the specific access (or even better use `.get()` if you expect the key to be optional). 

Also see #405

blueyed:
Yeah, that's a duplicate of https://github.com/neovim/pynvim/pull/405 unfortunately.

pynvim/api/common.py
blueyed: Had the same discussion in https://github.com/neovim/pynvim/pull/405#discussion_r315202545 already.

blueyed:
@noelevans 
Thanks for the PR - please review #405 instead.

PR #414, opened 2019-10-15
tests: remove cleanup_func

blueyed:
The "vim" fixture uses a new session always.

It was added when migrating to pytest in b65f62d, but is not really
necessary.  It could still be used to e.g. ensure tests are cleaning up
after themselves, but this is also not necessary really.

Adjusts existing tests, mainly with regard to handle IDs.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,6 +1,5 @@
 import json
 import os
-import textwrap
 
 import pytest
 
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -9,50 +9,6 @@ import pynvim
 pynvim.setup_logging("test")
 
 
-@pytest.fixture(autouse=True)
-def cleanup_func(vim):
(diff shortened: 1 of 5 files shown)


test/conftest.py
```diff
@@ -9,50 +8,6 @@
 pynvim.setup_logging("test")
 
 
-@pytest.fixture(autouse=True)
-def cleanup_func(vim):
```
justinmk: can we leave a stub here (with `pass`)?

blueyed: Why? :confused:

justinmk: if I want to add setup code later it will be a pain to find out how. Stubs don't hurt anything.

Not a blocker though.

blueyed: Stubs at least get called all the time (not really that bad, but still).
Should be discoverable through the Git history I think.
(and if you're used to pytest you would know what you need anyway)

PR #421, opened 2020-01-09
coveragerc: parallel=1

blueyed:
Fixes:

> INTERNALERROR> coverage.misc.CoverageException: Couldn't use data file '…/.coverage': UNIQUE constraint failed: meta.key

Ref: https://github.com/neovim/pynvim/pull/418#issuecomment-572617071


blueyed:
Scrutinizer complains about missing coverage:

> Code Coverage was not sent to Scrutinizer.
> Make sure that you have set-up your external service correctly.
> You can find more information on this in the documentation:
> 
>     https://scrutinizer-ci.com/docs/tools/external-code-coverage/

I guess that's what `ocular` is meant to do?  There's no error in logs currently, but needs investigation then.

justinmk:
Thanks for trying :)

justinmk:
Thanks for trying :) 

This is better than having all the travis builds fail, so let's merge it?

justinmk:
We could also remove scrutinzer, is it ever useful?

blueyed:
> We could also remove scrutinizer, is it ever useful?

Ok with me.  /cc @fwalch (via bfa672e)

I've checked that the number of runs matches our uploads / calls of `ocular`: https://github.com/neovim/python-client/blob/97ed85a8ca2deaa1689c0dd5b979a2d3ae72b716/.scrutinizer.yml#L131-L134.

blueyed:
Removed it.
Hopefully it stops being triggered once merged, otherwise somebody would need to disable it (https://scrutinizer-ci.com/g/neovim/pynvim/).

PR #423, opened 2020-01-14
asyncio: pipe_data_received: call _on_error

blueyed:
`_on_stderr` does not exist.

(found by mypy)

I might be missing something here, and it does not appear to be covered with tests.

### DIFF
--- a/pynvim/msgpack_rpc/event_loop/asyncio.py
+++ b/pynvim/msgpack_rpc/event_loop/asyncio.py
@@ -68,7 +68,7 @@ class AsyncioEventLoop(BaseEventLoop, asyncio.Protocol,
     def pipe_data_received(self, fd, data):
         """Used to signal `asyncio.SubprocessProtocol` of incoming data."""
         if fd == 2:  # stderr fd number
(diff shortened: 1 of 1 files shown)


pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -68,7 +68,7 @@ def pipe_connection_lost(self, fd, exc):
     def pipe_data_received(self, fd, data):
         """Used to signal `asyncio.SubprocessProtocol` of incoming data."""
         if fd == 2:  # stderr fd number
-            self._on_stderr(data)
+            self._on_error(data)
```
bfredl: `_on_error` is for a complete error message, not for streaming data. One could collect all the pieces and add the joined text in the `process_exited` call possibly. Or just change it to `pass` for now.

wookayin:
Actually at the moment, this line will be never called (until neovim's `--embed` RPC starts to write something on stderrs). Superseded by #543.

wookayin:
This was fixed by #543.

PR #430, opened 2020-02-05
imports vim module by default.

sunnyrahul25:
fixes neovim/neovim#11783.


justinmk:
Thanks. Perhaps add a test ? https://github.com/neovim/pynvim/blob/master/test/test_host.py

https://github.com/neovim/pynvim#development

PR #451, opened 2020-08-04
Close #450: Modify all imports in the repository to absolute imports

### LINKED ISSUE #450: Use absolute imports
Throughout the repository, relative imports are used.

As a result, simple and useful checks for the repository are missed.

For example:

`python3 -m doctest api/nvim.py`

Can't work because the internal imports are relative; however, if all the imports were refactored from

```python3

from .buffer import Buffer

# to

from pynvim.buffer import Buffer

```

simple checks like this would be possible.

farisachugthai:
@Shougo I realized that going ahead and changing all the imports wouldn't be very time intensive so I did.

In addition, I have verified that tests still pass with pytest in a venv.

However, I would like to note that:

`python -m doctest pynvim/msgpack_rpc/event_loop/base.py`

currently fails. There's a circular import issue, and even if that was resolved, it seems that Tracebacks were copy-pasted into `BaseEventLoop.__init__()`.

However that may be an issue for a different PR, and the imports mentioned in #450 are fixed

### DIFF
--- a/pynvim/__init__.py
+++ b/pynvim/__init__.py
@@ -6,13 +6,13 @@ import logging
 import os
 import sys
 
-from .api import Nvim, NvimError
-from .compat import IS_PYTHON3
-from .msgpack_rpc import (ErrorResponse, child_session, socket_session,
-                          stdio_session, tcp_session)
-from .plugin import (Host, autocmd, command, decode, encoding, function,
-                     plugin, rpc_export, shutdown_hook)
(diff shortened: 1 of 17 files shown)


farisachugthai:
I just noticed that Travis CI failed. After switching to absolute imports, the imports were in the incorrect order and flake8 failed as a result. Added 1 commit to fix this

justinmk:
Thanks for working on this!

PR #454, opened 2020-08-31
Doc updates: Indentation and underline length

farisachugthai:
Building the docs raised errors for me because the lines under the headers were too short.

In addition, a few of the code blocks weren't being properly highlighted (I.E. the bottom of <https://pynvim.readthedocs.io/en/latest/usage/remote-plugins.html> ) so I changed the lexer to signify that the code blocks are run inside of Vim and fixed the indentation and that seems to have worked for me.

### DIFF
--- a/docs/usage/remote-plugins.rst
+++ b/docs/usage/remote-plugins.rst
@@ -75,23 +75,26 @@ find your plugin. You can now invoke Neovim:
 
 Then run ``:UpdateRemotePlugins`` and your plugin should be activated.
 
-In case you run into some issues, you can list loaded plugins:
+In case you run into some issues, you can list your loaded plugins from inside
+Neovim by running ``:scriptnames`` like so.:
 
-.. code-block:: console
+.. code-block:: vim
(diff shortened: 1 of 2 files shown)


docs/usage/remote-plugins.rst
```diff
@@ -75,23 +75,26 @@ find your plugin. You can now invoke Neovim:
...
-.. code-block:: console
+.. code-block:: vim
 
     :set runtimepath
-      runtimepath=~/.config/nvim,/etc/xdg/nvim,~/.local/share/nvim/site,...,
+    runtimepath=~/.config/nvim,/etc/xdg/nvim,~/.local/share/nvim/site,...,
     ,~/g/path/to/your/plugin-git-repo
 
+    " Or alternatively
+    echo &rtp
```
justinmk, line 99: since line 94 has a colon let's use a colon here also, which is a nice convention that signals this is a vim command.

```suggestion
    :echo &rtp
```

PR #455, opened 2020-09-01
Reduce usage of deprecated imp module

farisachugthai:
Replace imp.new_module with types.ModuleType which is available from
py2.7 to 3+

### DIFF
--- a/pynvim/plugin/script_host.py
+++ b/pynvim/plugin/script_host.py
@@ -4,6 +4,7 @@ import io
 import logging
 import os
 import sys
+from types import ModuleType
 
 from pynvim.api import Nvim, walk
 from pynvim.compat import IS_PYTHON3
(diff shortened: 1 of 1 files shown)


justinmk:
Thank you!

farisachugthai:
Haha no problem! So I didn't get rid of every instance of imp because I'm not 100% sure how to replace the find_module and load_module calls.

2 of the classes with find/load_module calls are nested in a func themselves, so it'll probably be quite difficult to be sure that the API stays consistent before and after any changes.

Admittedly though, VimPathFinder has no `__init__` and 2 staticmethods so it might not be very difficult to completely remove.

Since it's inside of the path hook func I can't imagine anyone externally relying on that portion of the API

justinmk:
> So I didn't get rid of every instance of imp because I'm not 100% sure how to replace the find_module and load_module calls.

yeah, I tried awhile ago and only got so far, so was happy that you made a bit more progress :)

PR #458, opened 2020-09-14
Add metadata to setup.cfg for pypi

farisachugthai:
As it currently stands the repo's page on [pypi](https://pypi.org/project/pynvim/) doesn't include any metadata which would make it harder to find the project as well as displaying no information about it on the website.

This adds classifiers and keywords to help the project be easier to find as well as adding the README to it's description.

I also added some whitespace to one of the docstrings because building the docs were still generating a few errors, and a few simple additions to the gitignore that I needed while working on this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,31 @@
+[metadata]
+classifiers =
+    Environment :: Console
+    License :: OSI Approved :: Apache License
+    Intended Audience :: Developers
+    Intended Audience :: End Users/Desktop
+    Natural Language :: English
+    Operating System :: Android
+    Operating System :: Microsoft :: Windows :: Windows 10
+    Operating System :: POSIX
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+
+long_description = file: README.md, LICENSE
+long_description_content_type=text/markdown
+keywords = nvim, vim, pynvim
+platforms = any
+test_suite = test
+
+[options]
+include_package_data = True
+
 [aliases]
 test = pytest
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+tags
+.venv
+.pytest_cache/*
+.coverage*
 *.egg-info/
 __pycache__
 neovim/ui/screen.c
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -1,3 +1,31 @@
...
+    Operating System :: Android
+    Operating System :: Microsoft :: Windows :: Windows 10
+    Operating System :: POSIX:: Linux
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
bfredl, line 16: to be really picky we don't test for 3.9 in CI (yet)

setup.cfg
```diff
@@ -1,3 +1,31 @@
+[metadata]
+classifiers =
+    Environment :: Console
+    License :: OSI Approved :: Apache License
+    Intended Audience :: Developers
+    Natural Language :: English
+    Operating System :: Android
+    Operating System :: Microsoft :: Windows :: Windows 10
+    Operating System :: POSIX:: Linux
```
bfredl, line 9: Can we say something like `Operating System :: POSIX:: *` ? We aim to support all POSIXen that ppl want to use neovim on (including at least the BSD:s and macOS).

setup.cfg
```diff
@@ -1,3 +1,31 @@
+[metadata]
+classifiers =
+    Environment :: Console
+    License :: OSI Approved :: Apache License
+    Intended Audience :: Developers
```
bfredl, line 5: also end users I guess, as end users are expected to run `pip3 install pynvim` themselves.

setup.cfg
farisachugthai: Really picky is definitely preferable and I hugely appreciate you reviewing the PR!

Would it be preferable to remove this classifier or add 3.9 to CI? 3.9 is still the development branch for cpython but tmk it's available on Travis

setup.cfg
farisachugthai: Okay I can add `Intended Audience :: End Users/Desktop`

setup.cfg
farisachugthai: So unfortunately to show up in every subset of POSIX we'd have to explicitly list every single one. Alternatively, I'd imagine it would get the point across if the word Linux is simply removed.

bfredl: yes if `Operating System :: POSIX` works that should be enough.

setup.cfg
bfredl: remove for now. We can add it back later when we get into that :]

bfredl:
Why was this closed? It doesn't look resolved to me.

farisachugthai:
https://github.com/neovim/pynvim/pull/459

Reopened in a different pull request

PR #460, opened 2020-09-22
doc: Invoke pip via `python[23] -m pip` [ci skip]

e-kwsm:
https://github.com/pypa/pip/blob/38585adaecd88e386fe6fb434e37e4501a0cd45a/src/pip/_internal/utils/entrypoints.py#L10-L31

> WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ Install
 Supports python 2.7, and 3.4 or later.
 
 ```sh
-pip2 install pynvim
-pip3 install pynvim
+python2 -m pip install pynvim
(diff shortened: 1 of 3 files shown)


justinmk:
https://github.com/neovim/pynvim/commit/a08753422550840701d81746e642cf6531a70933

PR #461, opened 2020-10-02
Refactor: Remove usage of imp

farisachugthai:
Since there are no reasonable alternatives to imp in python2.7,
I figured it'd be the easiest to try importing python3 modules and if
that fails, fall back to the previous method. However this will remove
that annoying DeprecationWarning for everyone using python3.

### DIFF
--- a/pynvim/plugin/host.py
+++ b/pynvim/plugin/host.py
@@ -1,5 +1,4 @@
 """Implements a Nvim host for python plugins."""
-import imp
 import inspect
 import logging
 import os
@@ -9,7 +8,7 @@ from functools import partial
 from traceback import format_exc
 
 from pynvim.api import decode_if_bytes, walk
-from pynvim.compat import IS_PYTHON3, find_module
+from pynvim.compat import IS_PYTHON3
 from pynvim.msgpack_rpc import ErrorResponse
 from pynvim.plugin import script_host
 from pynvim.util import format_exc_skip, get_client_info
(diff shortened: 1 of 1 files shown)


bfredl:
don't be too afraid to break 2.7 if there is clear benefit. pynvim 0.5 can be py3.5+ only if we want to.

farisachugthai:
Hahaha thanks for the tip! I wanted this refactored to a separate function so that if the decision to go pure py3 gets made we have the weird py2 / py3 thing contained to 1 function and can simply remove the `except ImportError: import imp` malarky

Shougo:
I think it fixes my issue.
https://github.com/neovim/pynvim/issues/478

The PR is completed?

Shougo:
I think `compat.py` and `scripts_host.py` needs to be changed.

Shougo:
I have imported your commit in https://github.com/neovim/pynvim/pull/486.

pynvim/plugin/host.py
```diff
@@ -23,6 +22,27 @@
...
+    try:  # Python3
+        from importlib.util import module_from_spec, spec_from_file_location
+    except ImportError:  # Python2.7
+        import imp
+        from pynvim.compat import find_module
+        file, pathname, descr = find_module(name, [directory])
+        module = imp.load_module(name, file, pathname, descr)
+        return module
+    else:
+        spec = spec_from_file_location(name, location=directory)
```
wookayin, line 39: I don't think this is correct -- this seems to be always returning `None`, causing `ImportError`. `
```
spec_from_file_location("foobar", "/path/to/foobar.py") == ModuleSpec(name='foobar', ...)
spec_from_file_location("foobar", "/path/to/") == None
```

pynvim/plugin/host.py
```diff
@@ -161,8 +181,10 @@ def _load(self, plugins):
                     has_script = True
                 else:
                     directory, name = os.path.split(os.path.splitext(path)[0])
-                    file, pathname, descr = find_module(name, [directory])
-                    module = imp.load_module(name, file, pathname, descr)
+                    try:
+                        module = handle_import(directory, name)
+                    except ImportError:
+                        return
```
wookayin, line 187: pynvim remote plugin stopped working with handler/autocmd not found errors, but it was being silently ignored here

justinmk: Thanks for noticing that. Good thing we didn't release yet :D Any chance you can send a fix?

wookayin: I managed to fix this. I'm sending a PR very soon.

PR #468, opened 2020-11-25
test_invalid_utf8: Use distinct if/else branches

jamessan:
pytest isn't able to report the expected value when using the `X if Y else Z` syntax.

See pytest-dev/pytest#7727

### DIFF
--- a/test/test_buffer.py
+++ b/test/test_buffer.py
@@ -164,11 +164,19 @@ def test_mark(vim):
 def test_invalid_utf8(vim):
     vim.command('normal "=printf("%c", 0xFF)\np')
     assert vim.eval("char2nr(getline(1))") == 0xFF
-    assert vim.current.buffer[:] == ['\udcff'] if IS_PYTHON3 else ['\xff']
+
+    if IS_PYTHON3:
+        assert vim.current.buffer[:] == ['\udcff']
+    else:
+        assert vim.current.buffer[:] == ['\xff']
 
     vim.current.line += 'x'
-    assert vim.eval("getline(1)", decode=False) == '\udcffx' if IS_PYTHON3 else '\xffx'
-    assert vim.current.buffer[:] == ['\udcffx'] if IS_PYTHON3 else ['\xffx']
+    if IS_PYTHON3:
+        assert vim.eval("getline(1)", decode=False) == '\udcffx'
+        assert vim.current.buffer[:] == ['\udcffx']
+    else:
+        assert vim.eval("getline(1)", decode=False) == '\xffx'
+        assert vim.current.buffer[:] == ['\xffx']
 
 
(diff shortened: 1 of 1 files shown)


jamessan:
Forcing an error does show the expected reporting:

```
===================================================================================================== FAILURES =====================================================================================================
________________________________________________________________________________________________ test_invalid_utf8 _________________________________________________________________________________________________

vim = <pynvim.api.nvim.Nvim object at 0xb608d1f0>

    def test_invalid_utf8(vim):
        vim.command('normal "=printf("%c", 0xFF)\np')
        assert vim.eval("char2nr(getline(1))") == 0xFF
        if IS_PYTHON3:
            assert vim.current.buffer[:] == ['\udcff']
        else:
            assert vim.current.buffer[:] == ['\xff']
    
        vim.current.line += 'x'
        if IS_PYTHON3:
            assert vim.eval("getline(1)", decode=False) == '\udcffx'
>           assert vim.current.buffer[:] == ['\udcffxy']
E           AssertionError: assert ['\udcffx'] == ['\udcffxy']
E             At index 0 diff: '\udcffx' != '\udcffxy'
E             Full diff:
E             - ['\udcffx']
E             + ['\udcffxy']
E             ?          +

test/test_buffer.py:175: AssertionError
```

jamessan:
Guess I didn't need to force a failure to see how this looks.  Python 2 tests are failing and probably have been since they were added:

```
    def test_invalid_utf8(vim):
        vim.command('normal "=printf("%c", 0xFF)\np')
        assert vim.eval("char2nr(getline(1))") == 0xFF
    
        if IS_PYTHON3:
            assert vim.current.buffer[:] == ['\udcff']
        else:
>           assert vim.current.buffer[:] == ['\xff']
E           AssertionError: assert [''] == ['\xff']
E             At index 0 diff: u'' != '\xff'
E             Full diff:
E             - [u'']
E             + ['\xff']
```

blueyed:
Note that `assert 1 == 2 if 0 else 3` does not fail, but only `assert 1 == (2 if 0 else 3)`..!?

blueyed:
Distinct branches are good for code coverage though anyway.

jamessan:
I'm not sure what the correct behavior is supposed to be for Py2.  @Shougo or @bfredl, can you advise?

bfredl:
the correct default behaviour in python2 should be no decoding, i e get the raw `'\xFF'` byte out (and no `unicode` string)

jamessan:
> the correct default behaviour in python2 should be no decoding

Maybe you can push the right changes to my branch?  I'm not familiar with the pynvim APIs.

jamessan:
Closing in favor of #492, which rips out the py2 code, thus getting rid of the bug.

PR #471, opened 2020-12-15
Change the data type of a Buffer so that it works correctly

farisachugthai:
```python

from pynvim.api.buffer import *

```

Doesn't correctly import Buffer. `__all__` is incorrectly defined as a tuple and as a result it tries importing each individual letter instead of the actual class.

In addition, the python reference explicitly defines that `__all__` should be a list.

### DIFF
--- a/pynvim/api/buffer.py
+++ b/pynvim/api/buffer.py
@@ -3,7 +3,7 @@ from pynvim.api.common import Remote
 from pynvim.compat import IS_PYTHON3, check_async
 
 
-__all__ = ('Buffer')
+__all__ = ['Buffer']
 
 
 if IS_PYTHON3:


blueyed (approved):

pynvim/api/buffer.py
```diff
@@ -3,7 +3,7 @@
 from pynvim.compat import IS_PYTHON3, check_async
 
 
-__all__ = ('Buffer')
+__all__ = ['Buffer']
```
jamessan, line 6: Alternatively, make it an actual tuple by adding the missing comma:

```suggestion
__all__ = ('Buffer',)
```

Looks like this is needed in pynvim/api/{nvim,tabpage,window}.py, pynvim/msgpack_rpc/event_loop/__init__.py, and pynvim/plugin/host.py.

farisachugthai: This does technically work however according to the reference guide the data type of `__all__` is explicitly supposed to be a list

blueyed: @farisachugthai 
Ok.  A link/ref would have been useful, but I believe you.. :)

Can you update the other places also, please?

> Looks like this is needed in pynvim/api/{nvim,tabpage,window}.py, pynvim/msgpack_rpc/event_loop/init.py, and pynvim/plugin/host.py.

farisachugthai: Couldn't find it in the ref again but some googling got me this from the tutorial.

https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

The import statement uses the following convention: if a package’s `__init__.py` code defines a list named `__all__`, it is taken to be the **list** of module names that should be imported when `from package import *` is encountered. It is up to the package author to keep this list up-to-date when a new version of the package is released.

Yeah sure I can do that. Is it okay if we merge this and I do the rest in a separate PR?

wookayin:
This already has been fixed by #492 (ac03f5cbfe16), and released as a part of pynvim 0.5.

PR #477, opened 2021-01-28
Add nvim_input_mouse

glacambre:
This function was missing from the API :)


bfredl:
you can already do `nvim.api.input_mouse()` for any API function. New wrappers are only added if there is some more logic than just making the request.

glacambre:
Oh, sorry, I had missed that. Thanks for the advice! :)

PR #486, opened 2021-06-24
WIP: deprecate imp module

Shougo:
Since there are no reasonable alternatives to imp in python2.7,
I figured it'd be the easiest to try importing python3 modules and if
that fails, fall back to the previous method. However this will remove
that annoying DeprecationWarning for everyone using python3.

### DIFF
--- a/pynvim/compat.py
+++ b/pynvim/compat.py
@@ -2,37 +2,17 @@
 
 import sys
 import warnings
-from imp import find_module as original_find_module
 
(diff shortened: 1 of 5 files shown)


Shougo:
Can you test it?

Shougo:
No response...

justinmk:
Rebased in https://github.com/neovim/pynvim/pull/527

PR #492, opened 2021-07-26
Drop support for old python versions and add type annotations

### LINKED ISSUE #422: Drop support for Python 2 and 3.4 (type annotations)?
I'm looking into adding type annotations, and would like to do this for Python 3.5+ then.

Both Python 2 and 3.4 are EOL by now, and it might take time anyway to merge type annotations then anyway.

What do you think?

stevearc:
Fixes #422 

Dropping support for python 3.5 and earlier (3.5 reached end-of-life [September 5, 2020](https://www.python.org/downloads/release/python-3510/))

This adds coverage for most of the public API, but there are still some
areas that aren't covered yet.
* `script_host.py` is ignored entirely
* `host.py` is ignored entirely
* the `msgpack_rpc` submodule has partial coverage
* there are some Any annotations sprinkled around on some of the more
  complex functions

I recommend reviewing the commits in order

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,7 +1,7 @@
 Installation
 ============
 
-The Neovim Python client supports Python 3.6 or later.
+The Neovim Python client supports Python 3.7 or later.
 
 Using pip
 ---------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,25 +1,18 @@
 Installation
 ============
 
-The Neovim Python client supports Python 2.7, and 3.4 or later.
+The Neovim Python client supports Python 3.6 or later.
 
 Using pip
 ---------
 
 You can install the package without being root by adding the ``--user`` flag::
 
-    pip2 install --user pynvim
     pip3 install --user pynvim
-
-.. note::
-  
-    If you only use one of python2 or python3,
-    it is enough to install that version.
 
 If you follow Neovim HEAD, make sure to upgrade ``pynvim`` when you upgrade
 Neovim::
 
-    pip2 install --upgrade pynvim
     pip3 install --upgrade pynvim
 
 Install from source
@@ -32,5 +25,4 @@ Clone the repository somewhere on your disk and enter to the repository::
 
 Now you can install it on your system::
 
-    pip2 install .
     pip3 install .
--- a/pynvim/__init__.py
+++ b/pynvim/__init__.py
@@ -7,7 +7,6 @@ import os
 import sys
 
 from pynvim.api import Nvim, NvimError
-from pynvim.compat import IS_PYTHON3
 from pynvim.msgpack_rpc import (ErrorResponse, child_session, socket_session,
                                 stdio_session, tcp_session)
 from pynvim.plugin import (Host, autocmd, command, decode, encoding, function,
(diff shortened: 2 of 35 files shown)


bfredl:
In general I agree with this, but I would suggest for this PR to only drop obsolete python version support, and adding annotations to existing methods where we find that useful. We should consider more wrapper methods (and to what detail we want to specify those) in a separate follow-up PR.

stevearc:
Sounds good to me! I'll force push this PR to remove the extra wrapper methods

docs/installation.rst
```diff
@@ -1,25 +1,18 @@
 Installation
 ============
 
-The Neovim Python client supports Python 2.7, and 3.4 or later.
+The Neovim Python client supports Python 3.6 or later.
```
clason, line 4: 3.6 is end-of-life now, too (since Dec 23, 2021), so might as well bump to 3.7 (which still has till June 27, 2023).

stevearc: Bumped. Note that I left the `python_requires=">=3.6"` in the setup.py because of CI issues. The latest pypy3 in Ubuntu repos is only compatible with Python 3.6, so if we want to bump the `python_requires` we have to either stop testing on pypy3 (probably undesired) or wrangle the CI environment to install a more recent pypy3. I think that's outside of the scope of this PR.

We probably want to move to github workflows anyway, since travis-ci got rid of their easy-to-use free tier for open source projects.

stevearc: Yeah, looks like travis is already broken because it depends on https://github.com/neovim/bot-ci, which was archived.

cbrxyz:
This PR is awesome! Any way this can be reviewed/merged?

setup.py
```diff
@@ -45,6 +40,7 @@
       packages=['pynvim', 'pynvim.api', 'pynvim.msgpack_rpc',
                 'pynvim.msgpack_rpc.event_loop', 'pynvim.plugin',
                 'neovim', 'neovim.api'],
+      python_requires=">=3.6",
```
wookayin, line 43: This should be `>=3.7` as well

pynvim/__init__.py
```diff
@@ -77,8 +83,30 @@ def start_host(session=None):
...
+
+
+@overload
+def attach(session_type: Literal['stdio']) -> Nvim: ...
+
+
+def attach(
+    session_type: TTransportType,
+    address: str = None,
+    port: int = 7450,
```
wookayin, line 105: Is having a default argument here OK? Also, `str = None` is not a valid typing annotation. I think they might be `port: Optional[int] = None` or `address: Optional[str] = None`, etc.

wookayin (commented):
This is an amazing PR, I've left some comments at a couple of lines. 

One thing: I see a lot of `var: SomeType = None` instead of `var: Optional[SomeType] = None` overall in many places. Why not using `Optional`?

setup.py
stevearc: As mentioned in the comment above, this is because bumping the `python_requires` breaks the CI.

pynvim/__init__.py
stevearc: You are correct that the explicit `Optional[type]` is now the preferred style, [though it was not always such](https://peps.python.org/pep-0484/#union-types), and mypy I think still supports implicitly inferring the `Optional` (it did at the time of writing this PR, but it's been a while).

cbrxyz: The default argument does seem like a little bit of a weird change to be included in this PR... maybe something that was left in from testing/debugging?

benjamin-kirkbride:
Any update on getting this merged?

justinmk:
If someone can rebase this and test it out, we can merge it.

justinmk:
Rebased

PR #493, opened 2021-07-29
Add explicit API function definitions

stevearc:
Depends on #492 

With the addition of type annotations, there is now a reason to explicitly enumerate the API methods instead of relying on the magic `__getattr__`. If they are explicitly provided, plugin authors will get better autocompletion (via LSP) and more type safety guarantees.

Since this depends on #492, it has all of its commits. I'll rebase it once merged, but if anyone knows of a better Github workflow for dependent PR's, please let me know.

### DIFF
--- a/pynvim/api/__init__.py
+++ b/pynvim/api/__init__.py
@@ -8,8 +8,8 @@ from pynvim.api.buffer import Buffer
 from pynvim.api.common import decode_if_bytes, walk
 from pynvim.api.nvim import Nvim, NvimError
 from pynvim.api.tabpage import Tabpage
-from pynvim.api.window import Window
+from pynvim.api.window import Window, WindowConfig
 
 
 __all__ = ('Nvim', 'Buffer', 'Window', 'Tabpage', 'NvimError',
-           'decode_if_bytes', 'walk')
+           'decode_if_bytes', 'walk', 'WindowConfig')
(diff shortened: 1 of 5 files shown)


stevearc:
@justinmk That's a great idea! I'm happy to take a crack at auto-generation, or at least auto-verification. The PR this depends on is just collecting dust at the moment and I haven't seen much indication that it will be merged, so I'll probably defer the work until after there's more progress there.

justinmk:
> I'm happy to take a crack at auto-generation, or at least auto-verification

just auto-verification is enough to move forward. I would even say, let's wait on auto-generation until we get this first phase merged.

I think we're ready to move foward with https://github.com/neovim/pynvim/pull/492#issuecomment-889478664

justinmk:
Rebased after merging #492. 

Failing Windows CI is a known issue, unrelated to this PR.

justinmk:
Not in favor of this, for same reasoning as in https://github.com/neovim/node-client/pull/199#issuecomment-1628814410 : `pynvim` is a useful "reference implementation" Neovim client, but we need to be ruthless about keeping it extremely low-maintenance, because of the lack of ownership in the last 2+ years. 

Most APIs should be autogenerated at runtime, if at all. Meanwhile, `request('nvim_...')` is always available to clients.

PR #496, opened 2021-08-24
Fix the first call to sync functions returning null

### LINKED ISSUE #434: First call to function returns null
I'm using the standard example plugin:

```
import pynvim

@pynvim.plugin
class TestPlugin(object):

    def __init__(self, nvim):
        self.nvim = nvim

    @pynvim.function('TestFunction', sync=True)
    def testfunction(self, args):
        return 3

    @pynvim.command('TestCommand', nargs='*', range='')
    def testcommand(self, args, range):
        self.nvim.current.line = ('Command with args: {}, range: {}'
                                  .format(args, range))

    @pynvim.autocmd('BufEnter', pattern='*.py', eval='expand("<afile>")', sync=True)
    def on_bufenter(self, filename):
        self.nvim.out_write('testplugin is in ' + filename + '\n')

```

The first time I call `:echo TestFunction()`, it returns `v:null`. The second time returns the correct number. I also observed that if you open the plugin handler file first, then it will behave correctly and returns the correct number even for the first time.

As a side question, is there a way to reload remote plugins on the fly? I tried `:UpdateRemotePlugins` but it doesn't work. It's really annoying to restart nvim to see changes.

dmitmel:
This happens only right after instantiation of the plugin class, so calling any other asynchronous function seemingly fixes the problem.

Fixes #434.

### DIFF
--- a/pynvim/plugin/host.py
+++ b/pynvim/plugin/host.py
@@ -89,9 +89,9 @@ class Host(object):
         self._discover_functions(plugin, module_handlers, path, False)
 
         if sync:
-            self._request_handlers[name](*args)
+            return self._request_handlers[name](*args)
         else:
-            self._notification_handlers[name](*args)
+            return self._notification_handlers[name](*args)
 
     def _wrap_function(self, fn, sync, decode, nvim_bind, name, *args):
         if decode:


justinmk:
Nice, thank you! Does it look possible to add a test?

dmitmel:
Dunno, I don't see any tests involving plugin classes or similar. Plus this bug is just a one-off typo to me.

HiPhish:
Sorry for bumping this PR, but what are the chances of getting a bugfix release soon? I need this for a plugin I am writing currently. It's not urgent, but I would rather avoid following `HEAD` for just one plugin.

dmitmel:
As a workaround you can just post some dummy notification/request to the remote plugin when Vim loads. An empty method on the Python side is enough.

HiPhish:
I know, I already am doing that, so it's not a big deal, just ugly. (I am making a plugin which reads in the linter settings from pycodestyle and returns a dictionary, so I can set my Python file type settings automatically to match the linter settings)

wookayin:
A kind bump-up: pynvim v0.5.x (or any new version that includes this fix) is still unreleased.

justinmk:
I think @bfredl is required for that currently :)

bfredl:
What PyPI username should I add to the package?

justinmk:
"nvim" ? what was it before?

PR #506, opened 2022-05-09
Fix: evaluate v:true/v:false to python bool type

uzxmx:
This PR fixes an issue that when calling `vim.eval('v:true')`, or `vim.eval('v:false')`, it gives the wrong `str` type.

In the latest version of `vim`, when executing below Ex command, the result is `<class 'bool'>`. But in `neovim`, it is `<class 'str'>`.

```
:py3 print(type(vim.eval('v:true')))
```

### DIFF
--- a/pynvim/plugin/script_host.py
+++ b/pynvim/plugin/script_host.py
@@ -194,7 +194,7 @@ else:
 
 
 def num_to_str(obj):
-    if isinstance(obj, num_types):
+    if isinstance(obj, num_types) and not isinstance(obj, bool):
         return str(obj)
     else:
         return obj
--- a/test/test_host.py
+++ b/test/test_host.py
@@ -28,3 +28,10 @@ def test_host_async_error(vim):
     assert event[1] == 'nvim_error_event'
     assert 'rplugin-host: Async request caused an error:\nboom\n' \
            in h._on_error_event(None, 'boom')
+
+def test_legacy_vim_eval(vim):
+    h = ScriptHost(vim)
+    assert h.legacy_vim.eval('1') == '1'
+    assert h.legacy_vim.eval('v:null') == None
+    assert h.legacy_vim.eval('v:true') == True
+    assert h.legacy_vim.eval('v:false') == False


wookayin:
Follow-up: There was a concern raised in https://github.com/neovim/pynvim/issues/523#issuecomment-1495502011 that it doesn't conform with the `python-eval` doc: https://neovim.io/doc/user/if_pyth.html#python-eval

justinmk:
Can you send a PR to update the docs?

wookayin:
I think more decisions on the behavior of python-eval would be needed. Same as in the legacy vim, do we want to make the return type as string for numeric/boolean values? What's the main rationale behind any decision?

@uzxmx Do you have any idea why you wanted this change?

justinmk:
Reverted in https://github.com/neovim/pynvim/pull/562

See https://github.com/neovim/pynvim/issues/523#issuecomment-1495502011 : 

> the #506 PR is not compliant with the `vim.eval` [documentation](https://neovim.io/doc/user/if_pyth.html#python-eval).
> 
> The documentation states
> 
> > vim.eval(str)						python-eval
> > Evaluates the expression str using the vim internal expression
> > evaluator (see [expression](https://neovim.io/doc/user/eval.html#expression)).  Returns the expression result as:
> > (1) a string if the Vim expression evaluates to a string or number
> > (2) a list if the Vim expression evaluates to a Vim list
> > (3) a dictionary if the Vim expression evaluates to a Vim dictionary Dictionaries and lists are recursively expanded.
> 
> In Python, booleans are implemented as a [subclass of integers](https://docs.python.org/3/c-api/bool.html#:~:text=Booleans%20in%20Python%20are%20implemented%20as%20a%20subclass%20of%20integers.). In nvim, `v:true` and `v:false` evalutes to `true` and `false` (boolean values); for boolean values [vim uses numbers](https://neovim.io/doc/user/eval.html#Boolean). Therefore, by (1) and that booleans are integers in Python, `vim.eval('v:true')` should evaluate to the Python string `"True"` and not the boolean `True`. And therefore `:py3 print(type(vim.eval('v:true')))` should print out `<class 'str'>` and not `<class 'bool'>`. Thus, the extra check in #506 is redundant, right?

Issue #518, opened 2022-11-04
pynvim needs to be updated to use pyproject.toml


badboy1cdx:
When I just recently installed pynvim, I was presented with a deprecation warning that stated that starting in pip ver 23.1, pip will require compliance with using a pyproject.toml file for module installation, instead of the legacy setup.py file that pynvim currently uses.  Therefore I would recommend updating the package and migrating to a pyproject.toml file for setup as soon as possible to avoid having installation broken when pip starts forcing compliance.

justinmk:
any chance you could send a PR?

badboy1cdx:
I was actually considering doing just that.  I've just never worked with packaging before so I figured this would be a good learning opportunity.  I'll send the PR once I get it done.  I estimate two weeks due to current time constraints.

zachliu:
just saw the same msg today
```
DEPRECATION: pynvim is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```
@badboy1cdx thank you in advance!

orenteriab:
Bump on this as today I got this message:
```
Collecting pynvim
  Using cached pynvim-0.4.3.tar.gz (56 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
      [end of output]
```

badboy1cdx:
PR has been submitted.  See: #519

PR #519, opened 2022-11-18
Packaging update

badboy1cdx:
Hi all.  These changes are in response to issue #518 and bring the package to minimal compliance with packaging standards so we don't get deprecation warnings from pip due to outdated packaging format.  Highlights:

- Changed packaging and uploading commands in REAME to allow for building and uploading a source distribution as well as a wheel.
- Added language to setup.py to create universal wheel (necessary for continuing py2.7 compatibility - even if py2.7 is EOL).
- Add minimal pyproject.toml file to meet standards.

This is my first PR, so if I missed something please let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,8 @@ Release
    - bump the version in `pynvim/util.py` and `setup.py` (3 places in total)
 2. Make a release on GitHub with the same commit/version tag and copy the message.
 3. Run `scripts/disable_log_statements.sh`
-4. Run `python setup.py sdist`
+4. Run `python -m build`
     - diff the release tarball `dist/pynvim-x.y.z.tar.gz` against the previous one.
-5. Run `twine upload -r pypi dist/pynvim-x.y.z.tar.gz`
+5. Run `twine upload -r pypi dist/*`
     - Assumes you have a pypi account with permissions.
 6. Run `scripts/enable_log_statements.sh` or `git reset --hard` to restore the working dir.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -129,8 +129,8 @@ Release
    - bump the version in `pynvim/util.py` and `setup.py` (3 places in total)
 2. Make a release on GitHub with the same commit/version tag and copy the message.
 3. Run `scripts/disable_log_statements.sh`
-4. Run `python setup.py sdist`
+4. Run `python -m build`
```
wookayin, line 132: This seems to be only for a particular PEP517 choice [python-build](https://github.com/pypa/build). Shouldn't we have `pip install build` somewhere?

badboy1cdx: I hadn't thought of adding that, but will if you believe it's necessary.  I don't think it's really necessary for two reasons:

1. This is only for building the package for distribution - not for general end-user installation or re-installation.  The `build` package is not actually required for installing our package.
2. We're using the twine library a few lines below and don't mention `pip install twine` for that either.

I chose this method because it brings us into compliance with pep 517 with minimal effort and almost no change to the package as a whole.

At some point, someone _will_ probably want to do some more work to move a lot of the code from the `setup.py` file to the `pyproject.toml` file, but this at least makes us compliant and will stop pip from complaining every time someone installs the package (as it is doing now).

joe733 (approved):

justinmk:
Thank you for the help here! We definitely need more owners for pynvim, it's suffering from a lack of attention.

PR #524, opened 2023-04-27
This is a security test as part of a bug bouty, plese ignore.  jpjlnz

bazookaz1:
This is a security test as part of a bug bouty, plese ignore.  jpjlnz


bfredl:
Just tell us what university network we are supposed to ban from future contributions already.

Issue #528, opened 2023-07-14
Test failures: "Event loop is closed" with Python 3.12, Fedora Linux


michel-slm:
As tested using a scratch build on the Fedora build system: https://koji.fedoraproject.org/koji/taskinfo?taskID=103367353

I'm uploading the logs as otherwise they will get garbage collected within a couple of weeks.

[build.log](https://github.com/neovim/pynvim/files/12054071/build.log)
[root.log](https://github.com/neovim/pynvim/files/12054084/root.log)

Note: one error - the `AssertionError` in `test_buffer` - is introduced somewhere between 0.4.3 and the current master, it also fails when building for Fedora 38 which has Python 3.11, while building 0.4.3 succeeds and building 0.4.3 with the patches for handling `imp` removal passes tests on Fedora 38 and fails with the same `test_vim` event loop issue but passes `test_buffer`.

michel-slm:
Linking this here - not worth an issue of its own: this PR fixes warnings when using pytest 6+

https://github.com/neovim/pynvim/pull/529

michel-slm:
ah, the `test_buffer` issue is due to this https://github.com/neovim/pynvim/commit/82a2e14b96fbffa30caeefec6e86b668c96662eb

I guess unless using neovim nightly this will not work. I wonder if we can make it conditional on neovim version

michel-slm:
https://github.com/neovim/pynvim/pull/531 fixes `test_buffer` by not hardcoding the global value for 'default'

justinmk:
From your build.log file: 

```
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x7f4818b78400>
Traceback (most recent call last):
  File "/usr/lib64/python3.12/asyncio/base_subprocess.py", line 126, in __del__
    self.close()
  File "/usr/lib64/python3.12/asyncio/base_subprocess.py", line 104, in close
    proto.pipe.close()
  File "/usr/lib64/python3.12/asyncio/unix_events.py", line 768, in close
    self.write_eof()
  File "/usr/lib64/python3.12/asyncio/unix_events.py", line 754, in write_eof
    self._loop.call_soon(self._call_connection_lost, None)
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 772, in call_soon
    self._check_closed()
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 519, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
...
[...]

PR #534, opened 2023-09-22
fix: fix broken dynamic import of rplugin modules

wookayin:
The removal of `imp` package (#461) in order to supprot Python 3.12 had a bug where rplugins can't be loaded and the ImportError exception was silenced, making the remote provider throwing lots of errors. This commit fixes broken dynamic import of python modules from the registered rplugins.

We add tests for Host._load, with loading rplugins consisting of:

  (1) single-file module  (e.g., `rplugins/simple_plugin.py`)
  (2) package (e.g., `rplugins/mymodule/__init__.py`)

Note: As per https://docs.python.org/3.12/whatsnew/3.12.html#imp, I also tried something like:

```python
def _handle_import(path, name):
    # First try to import module
    spec = importlib.machinery.PathFinder.find_spec(name, [path])

    # or single-file package
    if spec is None:
        filename = os.path.join(path, name + '.py')
        loader = importlib.machinery.SourceFileLoader(name, filename)
        spec = importlib.util.spec_from_file_location(
            name, location=filename, loader=loader)

    if spec is not None:
        mod = importlib.util.module_from_spec(spec)
        spec.loader.exec_module(mod)
        return mod
    raise ImportError(name=name, path=path)
```
but this fails to resolve relative import within the package to unlike the previous implementation using `imp` (see the unit test fixtures). Possible more addition of `loader` would be needed. If you think this is the right way, please let me know so we can have additional work. Actually the remote host python should be able to import packages, so I don't see any reason why we couldn't use the approach of manipulating `sys.path` and import the module.

### DIFF
--- a/pynvim/plugin/host.py
+++ b/pynvim/plugin/host.py
@@ -1,10 +1,13 @@
 # type: ignore
 """Implements a Nvim host for python plugins."""
+
+import importlib
 import inspect
 import logging
 import os
 import os.path
 import re
+import sys
 from functools import partial
 from traceback import format_exc
 from typing import Any, Sequence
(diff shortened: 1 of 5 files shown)


wookayin:
/cc @justinmk @farisachugthai

[BTW CI fails on 3.12](https://github.com/wookayin/pynvim/actions/runs/6268481958) so I didn't include the commit https://github.com/wookayin/pynvim/compare/fix-import...wookayin:pynvim:ci-py312 in this PR. No idea why it fails to find the python3 host when python 3.12 is installed in the runner (py3.11 works fine). Possibly related to #528.

PR #535, opened 2023-09-23
ci: Add python3.12 support

### LINKED ISSUE #528: Test failures: "Event loop is closed" with Python 3.12, Fedora Linux

wookayin:
neovim's python3 provider detection has a bug on python3.12
(neovim/neovim#25316), so we work around this by explicitly setting
the `g:python3_host_prog` variable.

Also, pynvim on python3.12 requires greenlet 3.x.

Fixes #528

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,10 @@ extras_require = {
 
 if platform.python_implementation() != 'PyPy':
     # pypy already includes an implementation of the greenlet module
-    install_requires.append('greenlet')
+    if sys.version_info >= (3, 12):
+        install_requires.append('greenlet>=3.0.0rc3')
+    else:
+        install_requires.append('greenlet')
 
 if sys.version_info < (3, 8):
     install_requires.append('typing-extensions')
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         cache: 'pip'
         python-version: 3.11
     - name: install dependencies
-      run: pip install tox tox-gh-actions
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -24,7 +24,10 @@
 
 if platform.python_implementation() != 'PyPy':
     # pypy already includes an implementation of the greenlet module
-    install_requires.append('greenlet')
+    if sys.version_info >= (3, 12):
+        install_requires.append('greenlet>=3.0.0rc3')
+    else:
+        install_requires.append('greenlet')
```
justinmk, line 30: greenlet has been a pain, I wish we could eliminate it :/

justinmk:
Thank you!

Spirarel:
Getting this released would be great. As it stands, the readme is incorrect about supporting python 3.7 or later.

PR #540, opened 2023-10-13
packaging: add pynvim.__version__ attribute (0.5.0.dev0)

wookayin:
pynvim previously did not have the `pynvim.__version__` attribute,
which is a popular convention for python modules.

We add the `pynvim.__version__` string in addition to `pynvim.VERSION`,
and let `pynvim._version` be the only file that contains the version
metadata for the pynvim package. `setup.py` can also get the version
string from this file.

pynvim version is bumped up to 0.5.0.dev0 (from 0.4.3).

### DIFF
--- a/pynvim/_version.py
+++ b/pynvim/_version.py
@@ -1,0 +1,13 @@
+"""Specifies pynvim version."""
+# pylint: disable=consider-using-f-string
+
+from types import SimpleNamespace
+
+# see also setup.py
+VERSION = SimpleNamespace(major=0, minor=5, patch=0, prerelease="dev0")
+
+# e.g. "0.5.0", "0.5.0.dev0" (PEP-440)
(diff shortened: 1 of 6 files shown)


pynvim/_version.py
```diff
@@ -0,0 +1,13 @@
+"""Specifies pynvim version."""
+# pylint: disable=consider-using-f-string
+
+from types import SimpleNamespace
+
+# see also setup.py
+VERSION = SimpleNamespace(major=0, minor=5, patch=0, prerelease="dev0")
```
justinmk, line 7: Do we need to mention this in our release docs: https://github.com/neovim/pynvim#release

justinmk (approved):
Great choices!

pynvim/_version.py
wookayin: Oh I see, I missed that, will fix this.

Issue #505, opened 2022-04-11
OSError: EOF when trying to run example code on Windows


Toqozz:
I'm trying to run the example code on the main page of the README, namely:

```python
>>> from pynvim import attach
>>> nvim = attach('child', argv=["/bin/env", "nvim", "--embed", "--headless"])
```

But I get an error instead:

```python
>>> from pynvim import attach
>>> nvim = attach('child', argv=['C:\\Users\\Toqoz\\Documents\\code\\neovim-test\\Neovim2\\bin\\nvim.exe', '--embed', '--headless'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\__init__.py", line 122, in attach
    return Nvim.from_session(session).with_decode(decode)
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\api\nvim.py", line 80, in from_session
    channel_id, metadata = session.request(b'nvim_get_api_info')
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\msgpack_rpc\session.py", line 98, in request
    raise OSError('EOF')
OSError: EOF
```

Running the same thing from a script gives a bit more info, but I think it's irrelevant:
```
Traceback (most recent call last):
  File "C:\Users\Toqoz\Documents\code\neovim-test\something.py", line 4, in <module>
    nvim = attach('child', argv=['C:\\Users\\Toqoz\\Documents\\code\\neovim-test\\Neovim2\\bin\\nvim.exe', "--embed"])
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\__init__.py", line 122, in attach
    return Nvim.from_session(session).with_decode(decode)
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\api\nvim.py", line 80, in from_session
    channel_id, metadata = session.request(b'nvim_get_api_info')
  File "C:\Users\Toqoz\AppData\Roaming\Python\Python39\site-packages\pynvim\msgpack_rpc\session.py", line 98, in request
    raise OSError('EOF')
OSError: EOF
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x00000254C7607D30>
Traceback (most recent call last):
[...]

PR #542, opened 2023-10-13
ci: temporarily disable CI on windows

wookayin:
... until #505 is fixed.


justinmk:
Unless it's blocking us from releasing (or some other technical reason), I think we should keep this running as a reminder that it needs to be fixed.

wookayin:
Yes. Finally going to be fixed -- #543.

PR #544, opened 2023-10-16
test: add tests for attaching through socket, tcp, and stdio

wookayin:


wookayin:
on Windows: should we skip or xfail some tests?

- `stdio` fails. although implemented in `_connect_stdio`, it appears to have been broken for a while. 
- `socket` also fails. Seems we need to create socket files in a Windows-specific manner, but haven't figured out yet.

justinmk:
> on Windows: should we skip or xfail some tests?

xfail seems better, it will let us know when behavior is fixed.

wookayin:
Yes I agree. I rebased the tests. I don't have any working windows machines for development so it's quite difficult for me to fix the failing tests on windows. But will try to figure out sometime soon.

PR #543, opened 2023-10-15
Fix broken tests on Windows due to EOF error and other minor bugs

wookayin:
🎉✅ *TESTS on WINDOWS ARE NOW ALL GREEN* ✅🎉

Note: This PR consists of several commits and bugfixes, each of which is all required to fix the broken tests on windows platform for such a long time. I would like to request @justinmk that please merge this PR **without squashing** (either merge commit or rebasing), retaining all the individual commits.

## fix: EOF error on piped stderr being closed on Windows 

Most importantly, this PR should fix #505. Pipe (subprocess) based pynvim was broken since neovim 0.5.0, probably due to https://github.com/neovim/neovim/commit/c86d5fa981b0651167f789aaff685b42cad7aaca (neovim/neovim#11390): on Windows, stderr is closed for an embedded nvim.

See the commit message for more details.

## fix: do not leak resources across tests so as to prevent side effects

See the commit message for more details. asyncio event loops were not closing properly during tests, and were leaking in subsequent EventLoop, Session, Nvim instances.

PR #294, opened 2018-01-09
[WIP] Expose asyncio to python3 rplugins, removing python3.3 support

bfredl:
Currently, handling events beyond nvim itself is very limited, the only way additional sources can be added is with separate threads invoking `nvim.async_call`.  A problem is that the implemenation is dynamic between two event loops: asyncio and pyuv, so plugins can't use them.

The plan is to require asyncio on python3 so we can expose asyncio event loop to python3 rplugins. Python3.3 is hereby unsupported. Python2.7 rplugins is considered legacy: existing py2.7 rplugins remain supported for now, but new features might not be available (python2 via vim `:python` interface remain supported of course)

I know issues on windows has held up this change in the past (#243). We might need to change the asyncio backend on windows. IIRC there is even a libuv based backend, but I don't know it's status. I will look into this tomorrow.

I will make example and test this with subprocess communication, using msgpack to comminucate with a python helper process.  I know deoplete wants to use the feature for multiprocess. Ping @shougo

PR #545, opened 2023-10-16
Several refactoring on asyncio eventloop implementations

wookayin:
### refactor!: completely wipe out pyuv
  - No more pyuv, and improve docs. python's built-in asyncio is the only available backend (#294).

### fix: prevent closed pipe errors on closing asyncio transport

This follows up #543 and eliminates all the pytest warnings due to Proactor eventloop.
```
Example CI output (the pytest warnings that are going to be fixed):

Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x00000205288AD1C0>
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\proactor_events.py", line 116, in __del__
      _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\proactor_events.py", line 80, in __repr__
      info.append(f'fd={self._sock.fileno()}')
                        ^^^^^^^^^^^^^^^^^^^
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\windows_utils.py", line 102, in fileno
      raise ValueError("I/O operation on closed pipe")
  ValueError: I/O operation on closed pipe
  Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x00000205288AD1C0>
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\proactor_events.py", line 116, in __del__
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\proactor_events.py", line 80, in __repr__
    File "C:\hostedtoolcache\windows\Python\3.11.5\x64\Lib\asyncio\windows_utils.py", line 102, in fileno
  ValueError: I/O operation on closed pipe
```

## Other minor code improvements, dependent for the above changes

- refactor: expose event loop as a property rather than a field (see #294)
- refactor: use async coroutine and add debugging statements
- refactor: separate asyncio Protocol from AsyncioEventLoop
- refactor: improve typing

If you prefer breaking this down into several independent PRs, I can also do it -- please let me know.


wookayin:
@justinmk I wished this were also included in 0.5.0. But since this is just a matter of refactoring (which was necessary to figure out and fix a very long-standing bug #543) we could include this in later versions (but I'm not sure if it makes much sense to merge towards "minor patch" versions such as 0.5.1 rather than 0.6.0).

While I was working on some more drastic change where we completely remove the asyncio abstraction layer and greenlet (which is no longer necessary in modern python), this would be a good starting point so that I can continue refactoring.

justinmk:
> While I was working on some more drastic change where we completely remove the asyncio abstraction layer and greenlet (which is no longer necessary in modern python)

💯

PR #546, opened 2023-10-17
fix: invalid CLI use in scripts/{enable,disable}_log_statements.sh

wookayin:
- `find -name` => `find . -name`
- `xargs -i` => `xargs -I` (or can be replaced by the for loop)
- Use bash for loop instead of xargs to abort early when erorr happens
- Have one whitespace after `#` in the comments to conform with the
  standard python code style

(should the commit message be prefixed with `packaging:` or `fix:` ?)

### DIFF
--- a/scripts/disable_log_statements.sh
+++ b/scripts/disable_log_statements.sh
@@ -1,4 +1,9 @@
-#!/bin/sh -e
+#!/bin/bash
+
+set -e
 
 cd pynvim
-find -name '*.py' | xargs -i{} ../scripts/logging_statement_modifier.py {}
+for f in $(find . -name '*.py'); do
+    echo "Processing: $f"
+    ../scripts/logging_statement_modifier.py "$f"
+done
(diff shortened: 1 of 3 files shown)


scripts/disable_log_statements.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/sh -e
+#!/bin/bash
+
+set -e
 
 cd pynvim
-find -name '*.py' | xargs -i{} ../scripts/logging_statement_modifier.py {}
+for f in $(find . -name '*.py'); do
+    echo "Processing: $f"
+    ../scripts/logging_statement_modifier.py "$f"
```
justinmk, line 8: For reference: this script is a hack to workaround performance issues: https://github.com/neovim/pynvim/issues/250

Ideally we would avoid costly logging in a more architectural way rather than a build-time hack.

wookayin: Right, fully agreed; this seems really hacky. Just quick-fixing the shell script for now because it didn't work in macOS :)

PR #548, opened 2023-10-18
feat: Ex command `:py=` evaluate and print python expression 

wookayin:
The Ex command `:py`, `:python`, :`py3`, etc. can evaluate the line as an expression rather than a statement if the line starts with `=`, just like `:lua=`.

`:py= <expr>` is equivalent as `:py print(<expr>)`.

```vim
:py3= sys.version_info[:3]
:python3 =pynvim.__version__
```

## TODO

- How to document into `/doc/if_pyth.txt` (neovim core)? The feature would require pynvim 0.5.0 (or devel). It would make sense to submit a PR to neovim core only after a stable release of pynvim that contains this patch is released.

## Developer note:

~~should be merged (or rebased) after #547 to avoid conflicts~~ (Merged)

### DIFF
--- a/pynvim/plugin/script_host.py
+++ b/pynvim/plugin/script_host.py
@@ -82,7 +82,7 @@ class ScriptHost(object):
         """Handle the `python` ex command."""
         self._set_current_range(range_start, range_stop)
 
-        if script[0] == '=':
+        if script.startswith('='):
             # Handle ":py= ...". Evaluate as an expression and print.
             # (note: a valid python statement can't start with "=")
             expr = script[1:]
--- a/pynvim/plugin/script_host.py
+++ b/pynvim/plugin/script_host.py
@@ -81,10 +81,19 @@ class ScriptHost(object):
     def python_execute(self, script, range_start, range_stop):
         """Handle the `python` ex command."""
         self._set_current_range(range_start, range_stop)
+
+        if script[0] == '=':
+            # if starts with '=', evaluated as an expression and printed.
+            # (note: a valid python statement can't start with "=")
+            expr = script[1:]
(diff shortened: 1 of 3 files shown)


justinmk:
> How to document into /doc/if_pyth.txt (neovim core)? The feature would require pynvim 0.5.0 (or devel).

Better to get it in there asap. One more motivation for doing a pynvim release :)

pynvim/plugin/script_host.py
```diff
@@ -81,10 +81,19 @@ def teardown(self):
     def python_execute(self, script, range_start, range_stop):
         """Handle the `python` ex command."""
         self._set_current_range(range_start, range_stop)
+
+        if script[0] == '=':
+            # if starts with '=', evaluated as an expression and printed.
```
justinmk, line 86: a brief example is a high-bandwidth hint
```suggestion
            # Handle ":py= ...". Evaluate as an expression and print.
```

pynvim/plugin/script_host.py
```diff
@@ -82,7 +82,7 @@ def python_execute(self, script, range_start, range_stop):
         """Handle the `python` ex command."""
         self._set_current_range(range_start, range_stop)
 
-        if script[0] == '=':
+        if script.startswith('='):
```
wookayin, line 85: (script can't be an empty string here when called via rpc / `:python3` command, but it'd be a more error-safe code)

wookayin:
Hmm now I think we may want to use `repr(...)` to print. For example, `python3= "foo"` might print `'foo'` instead of `foo` only.

Need to make the behavior consistent with https://github.com/neovim/neovim/issues/25379

Issue #552, opened 2023-12-05
docs (readthedocs) CI is failing


justinmk:
The docs build https://readthedocs.org/projects/pynvim/builds/ is failing since 260a0b922bbd30be4ec32365faf37c2c70a7a4c4 . That commit triggered the build, but its code changes may be unrelated, instead we may need to update some dependencies/configuration: https://github.com/search?q=repo%3Aneovim%2Fpynvim+Sphinx&type=code

https://readthedocs.org/projects/pynvim/builds/22217681/

```
python -m sphinx -T -E -W --keep-going -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
Running Sphinx v7.2.6
making output directory... done

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/cmd/build.py", line 293, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/application.py", line 272, in __init__
    self._init_builder()
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/application.py", line 342, in _init_builder
    self.builder.init()
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/builders/html/__init__.py", line 219, in init
    self.init_templates()
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/builders/html/__init__.py", line 270, in init_templates
    self.theme = theme_factory.create(themename)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/pynvim/envs/latest/lib/python3.11/site-packages/sphinx/theming.py", line 230, in create
    raise ThemeError(__('no theme named %r found (missing theme.conf?)') % name)
sphinx.errors.ThemeError: no theme named 'sphinx_rtd_theme' found (missing theme.conf?)

Theme error:
[...]

PR #553, opened 2023-12-11
ci: fix failing readthedocs build by adding sphinx_rtd_theme dependency

wookayin:
This *might* fix the failing readthedocs CI #552 issue (can't fully test until merged).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -17,6 +17,8 @@ sphinx:
   fail_on_warning: true
 
 python:
-   install:
-   - method: pip
-     path: .
+  install:
+  - method: pip
+    path: .
+    extra_requirements:
+      - sphinx_rtd_theme


.readthedocs.yml
```diff
@@ -17,6 +17,8 @@ sphinx:
...
 
 python:
-   install:
-   - method: pip
-     path: .
+  install:
+  - method: pip
+    path: .
+    extra_requirements:
+      - sphinx_rtd_theme
```
justinmk, line 24: should it be `sphinx-rtd-theme` ?

wookayin: Actually this translated to `pip install .[sphinx_rtd_theme]` (sorry the docs were not clear...). So I think the solution is to change this with `docs` and add `sphinx-rtd-theme` into `extra_requires['docs']`.

PR #554, opened 2023-12-11
ci: fix failing readthedocs build

wookayin:
Ref #552.  Fixup to #553


justinmk:
Nice, thanks. Just noticed that the last succeeding build says:

    Running Sphinx v1.8.6

whereas the first failing build says:

    Running Sphinx v7.2.6

PR #556, opened 2023-12-11
fix(asyncio): close child_watcher before the main loop is closed

### LINKED ISSUE #555: Output flooded with EventLoop related messages
I have a Python script that is spawning thousands of short lived `nvim` instances.
With the latest release of pynvim  (i.e. `0.5.0`) the output gets flooded with the following messages: 

```
…
Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 40854 is closed
Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 40855 is closed
Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 40856 is closed
…
```

Is there a way to silence these messages ? A workaround, for the time being, is to downgrade to `0.4.3`.

Thank you for your time and the new release.

wookayin:
This might fix #555.

TODO:

- Reproduce the bug
- Add tests

### DIFF
--- a/pynvim/msgpack_rpc/event_loop/asyncio.py
+++ b/pynvim/msgpack_rpc/event_loop/asyncio.py
@@ -242,6 +242,11 @@ class AsyncioEventLoop(BaseEventLoop):
 
     @override
     def _close(self) -> None:
+
+        if self._child_watcher is not None:
+            self._child_watcher.close()
+            self._child_watcher = None
+
         def _close_transport(transport):
             transport.close()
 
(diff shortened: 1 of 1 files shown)


wookayin:
/cc @pierreganty can you please if this PR fixes your issue in #555?

To checkout a PR:

```bash
git clone https://github.com/neovim/pynvim
cd pynvim
git fetch origin pull/556/head
git checkout FETCH_HEAD
pip install -e .
```

and then run neovim (make sure you have a proper python3 host on which this devel version of pynvim is installed).

pierreganty:
I have tried and made sure it uses pynvim Version: 0.5.1.dev0 (I followed your clear instructions) but the behavior is the same: the output gets flooded with the same messages as described in #555.

wookayin:
Closing because this does not fix the issue. see #555

PR #562, opened 2024-03-16
Revert "fix: vim.eval('v:true') should return python bool #506"

### LINKED ISSUE #523: `num_to_str` boolean check makes `Ultisnips` snippet engine laggy
Hi 

Today I cloned the repository and installed the package directly instead from PyPi.

After reopening nvim I noticed that nvim was very laggy in all python files.
I found out that the issue was with my snippet engine `Ultisnips`.
Deleting the package and reinstalling it from PyPi made the lag go away.

The version on PyPi is 4.3 and so I went through the git history to check where something significant changed.
I located the issue to be this PR #506. 
This PR changes the `num_to_str` function to include a [boolean check](https://github.com/neovim/pynvim/blob/41edb05b035c53bc396bf11fb2815ab03f65a2fc/pynvim/plugin/script_host.py#L197).
The helper function is used in the `LegacyVim` object's `eval` function [here](https://github.com/neovim/pynvim/blob/41edb05b035c53bc396bf11fb2815ab03f65a2fc/pynvim/plugin/script_host.py#L203).
Browsing through the [Ultisnips](https://github.com/SirVer/ultisnips) source code reveals that it uses `vim.eval` at many places. 
Thus, this change is non-trivial in terms of Ultisnips performance.

I wonder what could be done to circument this issue?

samuelqp:
This reverts commit d549371b6fb647304cff5f88a71e2bc42d2ed30a. This fixes #523 and make UltiSnips usable. As mentioned by @jsr-p, the extra check is redundant and is not worth the slower performance.

### DIFF
--- a/test/test_host.py
+++ b/test/test_host.py
@@ -56,14 +56,3 @@ def test_host_async_error(vim):
     assert event[1] == 'nvim_error_event'
     assert 'rplugin-host: Async request caused an error:\nboom\n' \
            in h._on_error_event(None, 'boom')
-
-
-def test_legacy_vim_eval(vim):
-    h = ScriptHost(vim)
-    try:
-        assert h.legacy_vim.eval('1') == '1'
-        assert h.legacy_vim.eval('v:null') is None
-        assert h.legacy_vim.eval('v:true') is True
-        assert h.legacy_vim.eval('v:false') is False
-    finally:
-        h.teardown()
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ Pynvim defines some extensions over the vim python API:
 
 See the [Python Plugin API](http://pynvim.readthedocs.io/en/latest/usage/python-plugin-api.html) documentation for usage of this new functionality.
 
+### Known Issues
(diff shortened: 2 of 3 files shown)


samuelqp:
The checks are failing even without the changes that I made (when the fork is identical to the current upstream) (https://github.com/samuelqp/pynvim/actions/runs/8304465825).

justinmk:
Thanks for the PR! 

The tests are broken because they refer to a (internal) function that was removed in https://github.com/neovim/neovim/commit/eb5d15e3838f53e2fcd25989c88db87458e9f984
Fixed in https://github.com/neovim/pynvim/pull/563 , would you mind rebasing ?

samuelqp:
Thanks! Seems good now.

justinmk:
One more ask: can you add a "Known issues" section after [this section](https://github.com/neovim/pynvim?tab=readme-ov-file#python-plugin-api) which explains this quirk? So that we don't get repeat reports of https://github.com/neovim/pynvim/pull/506

test/test_host.py
```diff
@@ -56,14 +56,3 @@ def test_host_async_error(vim):
...
            in h._on_error_event(None, 'boom')
-
-
-def test_legacy_vim_eval(vim):
-    h = ScriptHost(vim)
-    try:
-        assert h.legacy_vim.eval('1') == '1'
-        assert h.legacy_vim.eval('v:null') is None
-        assert h.legacy_vim.eval('v:true') is True
-        assert h.legacy_vim.eval('v:false') is False
```
justinmk, line 67: Aren't these tests still useful? Could change the expected values.

PR #573, opened 2024-08-11
fix(tests): remove reference to pytest-runner

### LINKED ISSUE #572: pytest-runner deprecated
This project still uses pytest-runner, but the [upstream repository](https://github.com/pytest-dev/pytest-runner) is archived:
```
Deprecation Notice

pytest-runner depends on deprecated features of setuptools and relies on features that break security mechanisms in pip. For example 'setup_requires' and 'tests_require' bypass pip --require-hashes. See also https://github.com/pypa/setuptools/issues/1684.

It is recommended that you:

    Remove 'pytest-runner' from your setup_requires, preferably removing the setup_requires option.
    Remove 'pytest' and any other testing requirements from tests_require, preferably removing the tests_requires option.
    Select a tool to bootstrap and then run tests such as tox.
```

0-wiz-0:
This does not seem to be used any longer.

Closes #572.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,10 +13,9 @@ install_requires = [
 ]
 
 needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
-pytest_runner = ['pytest-runner'] if needs_pytest else []
 
 setup_requires = [
-] + pytest_runner
+]
 
 tests_require = [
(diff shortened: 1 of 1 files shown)


justinmk:
Seems like it, based on the git history. Thanks!

PR #579, opened 2024-10-31
fix(ci): macOS CI failing with python{3.7, 3.8, 3.9}

wookayin:
macos-latest now points to M1 runners (macos-14-arm64) which do not support python 3.7-3.9. For these versions, use macos-12 runner.

Note: please do not squash commits for this PR.


wookayin:
CI is green now, but let me do a little bit of cleanup later tody.

justinmk (approved):
Thanks! LMK when ready to merge.

PR #581, opened 2024-11-04
build: fix conditional `install_requires`

mgorny:
Fix conditional `install_requires` in `setup.py` to use environment markers instead of inline conditions.  The latter do not work correctly with universal wheels -- e.g. a wheel made on Python 3.12 would not have a dependency on `typing-extensions` at all, even if it were installed on Python 3.11 or older, and the other way around.


mgorny:
Thanks!

PR #584, opened 2025-01-03
fix: remove the use of asyncio.get_child_watcher() for python3.14

### LINKED ISSUE #583: broken on python3.14 ('asyncio' has no attribute 'get_child_watcher')
Hi,

unfortunatelly pynvim seems to be broken on 3.14

```python

  File "/Users/user/work/my/ylib/sandbox/py/nvim/t.py", line 2, in <module>
    nvim = pynvim.attach('child', argv=["/usr/bin/env", "nvim", "--embed", "--headless"])
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/__init__.py", line 144, in attach
    child_session(cast(List[str], argv)) if session_type == 'child' else
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/__init__.py", line 49, in child_session
    return session('child', argv)
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/__init__.py", line 23, in session
    loop = EventLoop(transport_type, *args, **kwargs)
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/event_loop/asyncio.py", line 132, in __init__
    super().__init__(transport_type, *args, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/event_loop/base.py", line 109, in __init__
    raise e
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/event_loop/base.py", line 106, in __init__
    getattr(self, '_connect_{}'.format(transport_type))(*args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/user/.local/venv/python3.14/lib/python3.14/site-packages/pynvim/msgpack_rpc/event_loop/asyncio.py", line 201, in _connect_child
    self._child_watcher = asyncio.get_child_watcher()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'asyncio' has no attribute 'get_child_watcher'




```

cryptomilk:
Use PidfdChildWatcher if os.pidfd_open is available and works, and otherwise use ThreadedChildWatcher which should work in any case.

Fixes #583

### DIFF
--- a/pynvim/msgpack_rpc/event_loop/asyncio.py
+++ b/pynvim/msgpack_rpc/event_loop/asyncio.py
@@ -188,10 +188,20 @@ class AsyncioEventLoop(BaseEventLoop):
 
     @override
     def _connect_child(self, argv: List[str]) -> None:
+        def get_child_watcher():
+            try:
+                return asyncio.get_child_watcher()
+            except AttributeError:  # Python 3.14
+                return None
+
+            return None
+
         if os.name != 'nt':
             # see #238, #241
-            self._child_watcher = asyncio.get_child_watcher()
-            self._child_watcher.attach_loop(self._loop)
+            watcher = get_child_watcher()
+            if watcher is not None:
+                watcher.attach_loop(self._loop)
+                self._child_watcher = watcher
 
         async def create_subprocess():
             transport: asyncio.SubprocessTransport  # type: ignore
@@ -250,7 +260,8 @@ class AsyncioEventLoop(BaseEventLoop):
             # Windows: for ProactorBasePipeTransport, close() doesn't take in
             # effect immediately (closing happens asynchronously inside the
             # event loop), need to wait a bit for completing graceful shutdown.
-            if os.name == 'nt' and hasattr(transport, '_sock'):
+            if (sys.version_info < (3, 13) and
+                    os.name == 'nt' and hasattr(transport, '_sock')):
                 async def wait_until_closed():
                     # pylint: disable-next=protected-access
                     while transport._sock is not None:


cryptomilk:
Looks like that the CI fails before it runs the tests.

justinmk:
expand this to see the test failures:

image

```
 FAILED test/test_buffer.py::test_options - OSError: EOF
  FAILED test/test_buffer.py::test_get_exceptions - OSError: EOF
  FAILED test/test_window.py::test_options - OSError: EOF
```

wookayin:
I will send a separate PR to fix a failure on python 3.13.

cryptomilk:
Oh, locally I'm running nvim 0.10.3 and everything works. GH runs nvim nightly?

cryptomilk:
I've rebased on master and it shows that the patch works fine with nvim-stable but fails with nvim-nightly.

justinmk:
windows ci shows an issue:

```
  C:\hostedtoolcache\windows\Python\3.13.1\x64\Lib\asyncio\unix_events.py:40: in <module>
      raise ImportError('Signals are not really supported on Windows')
  E   ImportError: Signals are not really supported on Windows
```

other than that, the failing `vim.current.buffer.options['invalid-option']` tests are a known issue 

> Note: Tests are still failing on nvim-nightly due to upstream bug, but as soon as [neovim/neovim#31894](https://github.com/neovim/neovim/issues/31894) is fixed CI should be back to green.

cryptomilk:
Maybe we need to make sure the ThreadedChildWatcher is used on Windoze?

pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -188,18 +193,33 @@ async def connect_stdout():
...
             pid = transport.get_pid()
             debug("child subprocess_exec successful, PID = %s", pid)
 
+            if os.name != 'nt':
+                # see #238, #241
+                pidfd_works = False
+                if PidfdChildWatcher is not None and hasattr(os, "pidfd_open"):
+                    try:
+                        fd = os.pidfd_open(pid)
+                    except Exception:
```
wookayin, line 209: Out of curiosity, under which settings can the `pidfd_open` syscall fail?

cryptomilk: It is a Linux syscall. So on Linux < 5.3 or BSD.

wookayin: On Linux < 5.3, `os.pidfd_open` won't exist at all. Precisely what exceptions are being caught, and when does it *fail* (while exists)?

cryptomilk: Invalid pid, pid already dead, not sure if a pid is a zombie. Are we sure transport.get_pid() only returned a valid working pid?

pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -188,18 +193,33 @@ async def connect_stdout():
...
+                        pass
+                    else:
+                        os.close(fd)
+                        pidfd_works = True
+
+                if pidfd_works:
+                    watcher = PidfdChildWatcher()
+                else:
+                    watcher = ThreadedChildWatcher()
```
wookayin, line 218: Just noting: ThreadedChildWatcher requires python 3.8+, PidfdChildWatcher requires python 3.9+. While we might want to drop support for python3.7, we could also simply fall back to `get_child_watcher` if neither exists.

I would suggest extracting these lines into an inner helper function (say `get_child_watcher`) and return either `PidfdChildWatcher`, `ThreadedChildWatcher`, or `asyncio.get_child_watcher()` as a fallback.

justinmk (approved):

cryptomilk:
I'm sick, it has to wait ...

justinmk:
If necessary, we can drop python 3.7 support.

justinmk:
https://github.com/neovim/neovim/issues/31894 was fixed, will re-run the tests here after a new nightly is posted.

cryptomilk:
Rebased, tests are mostly passing now. The Windows and MacOS tests with Python 3.13 are failing.

On Windows it looks like we have some closing workaround which fails with Python 3.13 now.
https://github.com/neovim/pynvim/blob/9c5ec674453a9742c141d201153bd12b73846355/pynvim/msgpack_rpc/event_loop/asyncio.py#L250

I have no idea what is wrong on MacOS.

cryptomilk:
I added a commit which disabled the workaround on Windows and that seems to work. Not sure if this is the right fix. I have no clue about Windows.

The failing test_attach.py::test_connect_stdioThe test on MacOS seems to be race condition. Now it works on 3.13 and fails on 3.12.

wookayin:
I don't think the macOS failure has something to do with this PR; it'll be a different bug that already existed. I'll see later if we can bring a proper fix.

justinmk:
> disabled the workaround on Windows and that seems to work

that was added in https://github.com/neovim/pynvim/commit/17fbcbccde9978a5caa05f571174bfff4a0b7bdd ; @wookayin do you see any risk with disabling it here?

justinmk (approved):

pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -188,10 +188,42 @@ async def connect_stdout():
...
+                os.close(fd)
+            except OSError:
+                return False
+
+            return True
+
+        def get_child_watcher():
+            if can_use_pidfd():
+                try:
+                    from asyncio.unix_events import PidfdChildWatcher
```
wookayin, line 209:
```suggestion
                try:
                    from asyncio import PidfdChildWatcher, ThreadedChildWatcher
                    return PidfdChildWatcher()
                except ImportError:
                    # FIXME Python 3.7, return None if we drop support
                    return asyncio.get_child_watcher()
                     
```

On python3.14 (macos), `asyncio.unix_events.PidfdChildWatcher` symbol does not exist; there is only `_PidfdChildWatcher`.

wookayin (changes requested):

pynvim/msgpack_rpc/event_loop/asyncio.py
cryptomilk: Are you sure that `can_use_pidfd()` returns `True` on MacOS?

pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -250,7 +282,8 @@ def _close_transport(transport):
             # Windows: for ProactorBasePipeTransport, close() doesn't take in
             # effect immediately (closing happens asynchronously inside the
             # event loop), need to wait a bit for completing graceful shutdown.
-            if os.name == 'nt' and hasattr(transport, '_sock'):
+            if (sys.version_info.major == 3 and sys.version_info.minor < 13 and
+                    os.name == 'nt' and hasattr(transport, '_sock')):
```
wookayin, line 286:
```suggestion
            if sys.version_info < (3, 13) and os.name == 'nt' and hasattr(transport, '_sock')):
```

(though python 4.13 is too far in the future..)

In replying to https://github.com/neovim/pynvim/pull/584#issuecomment-2583355719: I haven't tested on Windows + python3.13, but I remember for older python versions this is still necessary.

wookayin:
I cherry-picked the commit eb14e83 for bumping up msgpack as it's a bit tangential to this fix.

Regarding the workaround in 17fbcbccde9978a5caa05f571174bfff4a0b7bdd, we can try disabling the workaround for python3.13 on Windows to see if the bug has been fixed. But I believe for older versions this workaround will still be necessary.

pynvim/msgpack_rpc/event_loop/asyncio.py
wookayin: Actually no. Though symbols can be imported, I agree they won't work.

```python
>>> from asyncio.unix_events import can_use_pidfd
>>> can_use_pidfd()
False
>>> from asyncio import PidfdChildWatcher
ImportError: cannot import name 'PidfdChildWatcher' from 'asyncio'

>>> from asyncio.unix_events import _PidfdChildWatcher
>>> _PidfdChildWatcher()
<asyncio.unix_events._PidfdChildWatcher object at 0x105d11010>

>>> from asyncio.unix_events import _ThreadedChildWatcher
>>> _ThreadedChildWatcher()
<asyncio.unix_events._ThreadedChildWatcher object at 0x105d10c20>
```
and same for `ThreadedChildWatcher`. In the source code of python3.14 alpha distribution for macOS, those watcher classes are prefixed in `_`. So I think `asyncio.unix_events` is fine but platform-dependent. 

On my macOS python3.14 env, it falled back to `asyncio.get_child_watcher()` and error again (it's gone). It seems that our current fix works for Linux, but I'm not yet sure about macOS.

pynvim/msgpack_rpc/event_loop/asyncio.py
cryptomilk: I added this. Thanks.

pynvim/msgpack_rpc/event_loop/asyncio.py
cryptomilk: Ah, I see. Pushed a fix.

wookayin: Thanks. I know it's not easy to reproduce from your end, but somehow  `asyncio.ThreadedChildWatcher` does not exist in python3.14. So it still falls back to `asyncio.get_child_watcher()` :/  I guess this is a regression in python3.14 dev versions --- because this works in python3.13 for me.

Since python3.14 is not a stable/released version, this shouldn't be a major blocker.

cryptomilk:
I really start to hate Python.

https://github.com/python/cpython/commit/9d2e1ea3862e5950d48b45ac57995a206e33f38b
https://github.com/python/cpython/issues/120804

Every 2 versions they completely break the API. WTF.

cryptomilk:
Looks like supporting 3.14 is a much bigger rewrite.

wookayin:
I appreciate your patience. Now as I think about the role of child_watcher, it's not quite crucial part of the pynvim functionality and I think we actually don't need it other than preventing resource leak when the event loop closes.

Let's make our life simpler and easier: my proposal is to simply just disable child_watcher in python3.14 (i.e. set `child_watcher = None`), and let pynvim just work for python3.14.

cryptomilk:
Looks like I misunderstood how things are now. I hope this patch fixes it now.

pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -188,10 +188,19 @@ async def connect_stdout():
...
+        def get_child_watcher():
+            try:
+                return asyncio.get_child_watcher()
+            except AttributeError:  # Python 3.14
+                return None
+
+            return None
+
         if os.name != 'nt':
-            # see #238, #241
```
wookayin, line 192: Let's keep this comment?

wookayin (approved):

pynvim/msgpack_rpc/event_loop/asyncio.py
cryptomilk: I added it back.

justinmk:
Thank you both!

PR #587, opened 2025-01-08
Microoptimize the walk() function

### LINKED ISSUE #250: performance issues
I'm profiling/debugging performance bottlenecks in the python client.

Two things stuck out so far.

- ~~Calls to BufferedFormatter in logging, when logging was disabled. This can be fixed by gating any `debug()`, `warn()` etc calls behind an `if _debug` statement as shown in asyncio.~~ 
  - (**Edit:** fixed, see below)
- `walk(to_nvim)` and `walk(from_nvim)` are extremely expensive calls on nested object trees. For example, if I have a 5000 line buffer and get/set the entire thing, these calls seem to account for at least 80% of the 20ms spent. I fixed this by adding the following functions on the `Remote` and `Nvim` classes and invoking it for `buf_get_lines` and `buf_set_lines` instead of request_raw:

```
class Remote:
    def request_raw(self, name, *args, **kwargs):
        return self._session.request_raw(name, self._session._to_nvim(self), *args, **kwargs)

class Nvim:
    def request_raw(self, name, *args, **kwargs):
        return self._session.request(name, *args, **kwargs)
```

Profiling data generated with yappi on Python 3.3 and my test script are attached.
[after.txt](https://github.com/neovim/python-client/files/811433/after.txt)
[before.txt](https://github.com/neovim/python-client/files/811432/before.txt)

```
import neovim

sample = ['a'] * 10000

def setup():
    n = neovim.attach('child', argv=['/Users/aegis/build/neovim/build/bin/nvim', '--embed'])
    buf = n.current.buffer
    buf[:] = sample
    return buf

def bufeq(buf):
    buf[:] = sample

def eqbuf(buf):
    _ = buf[:]

buf = setup()
for i in range(10):
    bufeq(buf)

for i in range(10):
    eqbuf(buf)

import sys
sys.exceptionhandler = lambda *a: 0
```

kmod:
I noticed that it takes pynvim about 4ms to attach to an nvim instance for me, and 3ms of that is due to the single line:
  metadata = walk(decode_if_bytes, metadata)

This commit reduces the walk() time down to 1.5ms, which brings the total attach time down to 2.5ms. This is helpful for me because in my use case I end up connecting to all of the currently-running nvim processes and this starts to take a noticeable amount of time. Unfortunately parallelization does not help here due to the nature of the slowness.

walk() is expensive because it does a very large amount of pure-python manipulation, so this commit is just some tweaks to reduce the overheads:

- *args and **kw make the function call slow, and we can avoid needing them by pre-packing the args into fn via functools.partial
- The comprehensions can be written to directly construct the objects rather than create a generator which is passed to a constructor
- The typechecking is microoptimized by calling type() once and unrolling the `type_ in [list, tuple]` check

I did notice that in my setup the metadata contains no byte objects, so the entire call is a noop. I'm not sure if that is something that could be relied on or detected, which could be an even bigger speedup.

fix #250

### DIFF
--- a/pynvim/api/common.py
+++ b/pynvim/api/common.py
@@ -241,11 +241,15 @@ def decode_if_bytes(obj: T, mode: TDecodeMode = True) -> Union[T, str]:
     return obj
 
 
-def walk(fn: Callable[..., Any], obj: Any, *args: Any, **kwargs: Any) -> Any:
-    """Recursively walk an object graph applying `fn`/`args` to objects."""
-    if type(obj) in [list, tuple]:
-        return list(walk(fn, o, *args) for o in obj)
-    if type(obj) is dict:
-        return dict((walk(fn, k, *args), walk(fn, v, *args)) for k, v in
-                    obj.items())
-    return fn(obj, *args, **kwargs)
+def walk(fn: Callable[[Any], Any], obj: Any) -> Any:
+    """Recursively walk an object graph applying `fn` to objects."""
+
+    # Note: this function is very hot, so it is worth being careful
+    # about performance.
+    type_ = type(obj)
+
+    if type_ is list or type_ is tuple:
+        return [walk(fn, o) for o in obj]
+    if type_ is dict:
+        return {walk(fn, k): walk(fn, v) for k, v in obj.items()}
+    return fn(obj)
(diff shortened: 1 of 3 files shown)


kmod:
I believe this is the same issue as was talked about in #250

justinmk (approved):
Nice, thanks!

justinmk:
Note that CI looks yucky right now, we're working on that :)

kmod:
Haha yeah at first I got worried about how I broke so much without realizing it

Also wanted to say, thanks for the project! It's been very handy

PR #589, opened 2025-01-10
fix: improve EOFError exception when remote nvim crashes

wookayin:
Problem: When the remote Nvim instance is aborted or terminates, the pynvim client will face an exception `OSError: EOF`. It is not very clear what is wrong and why EOF is received.

Solution: This happens when the remote the nvim process terminates unexpectedly (or a tcp/file socket is broken or gets disconnected). We can provide a bit more detailed information about why the asyncio session is stopping, through asyncio Protocol. An `EOFError` will be raised instead of OSError.

For example, during pynvim's unit tests we may see:
```
    EOFError: process_exited: pid = 40000, return_code = -6
```
which means that the Nvim subprocess (pid = 40000) exited unexpectedly after getting SIGABRT (signal 6), i.e. assertion failure. Other error messages such as SIGSEGV (signal 11) or connection lost (through socket) are also possible.

### DIFF
--- a/pynvim/msgpack_rpc/event_loop/asyncio.py
+++ b/pynvim/msgpack_rpc/event_loop/asyncio.py
@@ -47,13 +47,14 @@ class Protocol(asyncio.Protocol, asyncio.SubprocessProtocol):
     @override
     def connection_made(self, transport):
         """Used to signal `asyncio.Protocol` of a successful connection."""
-        del transport  # no-op
+        self._transport = transport
 
     @override
     def connection_lost(self, exc: Optional[Exception]) -> None:
         """Used to signal `asyncio.Protocol` of a lost connection."""
-        debug(f"connection_lost: exc = {exc}")
-        self._on_error(exc if exc else EOFError())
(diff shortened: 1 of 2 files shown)


pynvim/msgpack_rpc/event_loop/asyncio.py
```diff
@@ -250,11 +269,13 @@ def _close_transport(transport):
             # Windows: for ProactorBasePipeTransport, close() doesn't take in
             # effect immediately (closing happens asynchronously inside the
             # event loop), need to wait a bit for completing graceful shutdown.
-            if os.name == 'nt' and hasattr(transport, '_sock'):
+            if os.name == 'nt' and hasattr(transport, "_sock"):
```
justinmk, line 272: doesn't this project use single quotes?

justinmk (approved):
Extremely useful!

wookayin:
Sorry for my recent inactivities. I've rebased and resolved the conflict, so please feel free to merge this one.

PR #590, opened 2025-01-30
ci: neovim nightly name was changed

justinmk:
see https://github.com/neovim/neovim/pull/32194


justinmk:
`test_connect_stdio` now fails on python 3.13 macOS 😕

PR #615, opened 2026-04-09
fix(tests):  handle both neovim 0.12 invalid cursor error messages

### LINKED ISSUE #614: test_command_error fails with Neovim 0.12.x due to changed error message
Hi there,

I noticed that `test_command_error` in `test/test_vim.py` fails when running against Neovim 0.12.0+. 

The test asserts the exact error message string "Cursor position outside buffer", but this message was changed to "Invalid cursor line: out of range" in Neovim 0.12.0 (specifically in neovim commit [680d25e](https://github.com/neovim/neovim/commit/680d25e5b3c1efba13bd8f5c0bd6bb5ab431361f)).

Not sure what the right thing to do is, but I guess the test should probably be a bit less fragile.

The only reason I noticed was because i was attempting to update the neovim package to 0.12.1 in the guix package manager, which runs a packages test suite as part of its build by default...

campbellr:
Fixes #614 

In 0.12.0, the error message is now:

```
Invalid cursor line: out of range
```

This updates the test to handle both old and new error messages.

### DIFF
--- a/test/test_vim.py
+++ b/test/test_vim.py
@@ -1,6 +1,7 @@
 """Tests interaction with neovim via Nvim API (with child process)."""
 
 import os
+import re
 import sys
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks! 

The failing `test_connect_stdio` test is a known issue, unrelated to this pr.

PR #621, opened 2026-09-17
fix: scope vim.windows to the current tabpage

### LINKED ISSUE #592: vim.windows behaviour inconsistent between Vim and NVim
To reproduce, open vim and create a second tab, then print the list of all `vim.windows`.

```
$ vim # or nvvim
:tabedit
:python3 print(list(vim.windows))
```

Under Vim this only returns the windows in the currently visible buffer, i.e. ```[<window 0>]```. Under Neovim this returns all windows in all buffers, i.e. ```[<Window(handle=1000)>, <Window(handle=1001)>]```. This is inconsistent behavior to Vim. 

This yields problems when you for example iterate through all windows and try to do actions within them, like so:

```
    def _close_all():
        for idx,win in enumerate(vim.windows, 1):
            vim.command("%iwincmd w" % idx)
            if "nofile" in vim.eval("&buftype"):
                vim.command(":bwipeout")
                return _close_all()
```

Shubham-Padkonde:
Fixes #592.

After opening a second tab, `vim.windows` currently includes windows from both tabs. This differs from Vim's Python interface and can make plugins act on windows outside the current tab.

Use `nvim_tabpage_list_wins(0)` for the remote sequence. Resolving tabpage 0 on each request also makes a saved reference to `vim.windows` follow tab switches and tab closure. Explicit `tabpage.windows` remains bound to its own tab. `RemoteSequence` accepts optional request arguments to support this without adding an extra round trip.

The regression exercises a real embedded Neovim instance, including iteration, indexing, slicing, membership, switching and closing tabs. Existing cross-tab tests now address the intended tab explicitly. The documentation points callers that need every tab's windows to `vim.api.list_wins()`.

Validation on Linux, Python 3.12, Neovim 0.12.5:
- New regression fails on unchanged upstream code and passes with the fix.
- Full pytest suite: 81 passed.
- Full flake8 and mypy checks pass (39 source files).
- Sphinx HTML documentation build passes.

This changes the previously incorrect all-tabs behavior of `vim.windows`; callers intentionally depending on that behavior should use `vim.api.list_wins()`. Windows/macOS and Neovim's separate upstream test suite were not run locally.

Prepared with AI assistance; the change and test results were checked against the repository.


justinmk:
Looks like slop , based on the "Validation" section. I didn't read further. Please read https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md

